PR #8, opened 2018-11-07
Stub landing and authentication pages

coderberry:


hopsoft:
Are we using both MailChimp and MailGun. I noticed that both of these gems are included in the Gemfile.

```ruby
gem "gibbon" # MailChimp
gem "mailgun-ruby"
```

hopsoft:
Do we need 2 solutions for email HTML formatting? We have both of these gems as dependencies.

```ruby
gem "inky-rb" # zurb foundation based email formatting
gem "premailer-rails"
```

PR #31, opened 2018-11-14
User Management pages

coderberry:
#17


app/controllers/invitations_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+      else
+        set_flash_message :notice, :updated_not_active if is_flashing_format?
+        respond_with resource, location: new_session_path(resource_name)
+      end
+    else
+      resource.invitation_token = raw_invitation_token
+      respond_with_navigational(resource) { render :edit }
+    end
+  end
+end
```
hopsoft: Looks like we're overriding some devise stuff? I wonder if we could call `super` rather than duplicate.

app/views/@shared/_page_heading.html.erb
```diff
@@ -129,11 +133,23 @@
                   </li>
                   <!-- Accessibility -->
 
-                  <!-- Others -->
-                  <li class="nav-item u-header__nav-item">
-                    <a class="nav-link u-header__nav-link" href="plans.html">Plans</a>
-                  </li>
-                  <!-- Others -->
+                  <% if current_user.administrator? %>
```
hopsoft: We'll likely want to swap these types of checks out for the `authorizable` pattern on the images branch after it merges.

PR #40, opened 2018-11-16
Properties CRUD

coderberry:


app/controllers/concerns/sortable.rb
```diff
@@ -0,0 +1,23 @@
...
+  def order_by
+    "#{sort_column} #{sort_direction}"
+  end
+
+  def sort_column
+    return params[:column] if sortable_columns.include?(params[:column])
+    return "name"          if sortable_columns.include?("name")
+    "created_at"
+  end
+
```
hopsoft: I would define a default implementation for `sortable_columns` with a note to override in including controllers.

```ruby
# Abstract method that should be overridden in including controllers
def sortable_columns
  []
end
```

hopsoft: You could even raise `NotImplementedError` with the message.

app/controllers/properties_controller.rb
```diff
@@ -24,7 +28,7 @@ def show
 
   # GET /properties/new
   def new
-    @property = Property.new
+    @property = Property.new(user_id: current_user.id, status: "pending")
```
hopsoft: Might want to update to `current_user.properties.build` for consistency. I think that's a pattern that should be established everywhere its appropriate.

app/helpers/application_helper.rb
```diff
@@ -115,4 +115,37 @@ def noty_flash
...
+          #{title}
+          <div class="ml-2">
+            #{link_up}
+            #{link_down}
+          </div>
+        </div>
+      </th>
+    EOS
+
+    tr.html_safe
```
hopsoft: I would prefer a real template under `@shared/_sortable_tr.html.erb`. This helper method would render the partial rather then using inline HTML.

app/helpers/properties_helper.rb
```diff
@@ -2,20 +2,30 @@
 
 module PropertiesHelper
   def property_types_for_select
-    ENUMS::PROPERTY_TYPES.values
+    ENUMS::PROPERTY_TYPES.values.map { |val| [val.humanize, val] }
```
hopsoft: 💯

app/helpers/properties_helper.rb
```diff
@@ -2,20 +2,30 @@
...
+  end
+
+  def property_status_color(status)
+    {
+      "pending" => "warning",
+      "active" => "success",
+      "rejected" => "danger",
+      "archived" => "muted",
+      "blacklisted" => "dark"
+    }[status]
```
hopsoft: This should probably move to our `ENUMS` implementation. Thoughts?

app/views/property_searches/_form.html.erb
```diff
@@ -1,34 +1,16 @@
...
-            <%= tag.div f.input(:statuses, collection: property_statuses_for_select, required: false, input_html: { multiple: true, data: { controller: "select" }}), class: "col-12 col-md-auto" %>
-            <%= tag.div f.input(:ad_templates, collection: ad_templates_for_select, required: false, input_html: { multiple: true, data: { controller: "select" }}), class: "col-12 col-md-auto" %>
-            <%= tag.div f.input(:languages, collection: languages_for_select, required: false, input_html: { multiple: true, data: { controller: "select" }}), class: "col-12 col-md-auto" %>
-            <%= tag.div f.input(:keywords, collection: keywords_for_select, label: "Topics", required: false, input_html: { multiple: true, data: { controller: "select" }}), class: "col-12 col-md-auto" %>
+<%= simple_form_for @property_search, wrapper: :filter_form do |f| %>
+    <%= f.input(:name, required: false) %>
+    
+    <%# TODO: Is this something we should cache or have elsewhere? %>
+    <%= f.input(:user_id, label: "Publisher", collection: User.publisher.order(last_name: :asc).map { |u| [u.name, u.id] }, required: false, input_html: { data: { controller: "select" }}) %>
+    <%# ENDTODO %>
```
hopsoft: We'll probably want to cache this. but we can punt for now I think. We may also want to restrict it to `administrators` only.

vendor/theme/include/scss/_user.scss
```diff
@@ -1,3 +1,40 @@
...
+    border-radius: 0.25rem !important;
+    border: none !important;
+    padding: .25rem .75rem !important;
+    margin: 0 0 .25rem .25rem !important;
+
+    .select2-selection__choice__remove {
+      margin-right: 6px !important;
+    }
+  }
+}
```
hopsoft: I think we should move our custom CSS out of the vendor dir and under `app/javascripts/src/app/stylesheets`. Thoughts?

PR #53, opened 2018-11-19
Campaigns CRUD

coderberry:


app/controllers/campaigns_controller.rb
```diff
@@ -92,6 +94,25 @@ def set_user
...
+      :daily_budget,
+      :total_budget,
+      countries: [],
+      keywords: [],
+      negative_keywords: []
+    )
+  end
+
+  def sortable_columns
+    %w[name end_date total_budget_cents status created_at user.first_name]
```
hopsoft: Would prefer to see this split over multiple lines and sorted. Much easier to scan.

app/models/campaign.rb
```diff
@@ -123,6 +124,12 @@ def date_range
     "#{start_date.to_s "mm/dd/yyyy"} #{end_date.to_s "mm/dd/yyyy"}"
   end
 
+  def date_range=(value)
+    dates = value.split(" - ")
+    self.start_date = Date.strptime(dates[0], "%m/%d/%Y")
+    self.end_date   = Date.strptime(dates[1], "%m/%d/%Y")
+  end
```
hopsoft: Would prefer to see this use `to_s` with one of the `config/initializers/date_formats.rb`

coderberry: This is parsing, not printing

PR #55, opened 2018-11-20
Tweaking seeds

hopsoft:
Seed with default of 10k impressions in <= 15 secs.

Seed with 30M impressions in <= 40 mins.


db/seeds.rb
```diff
@@ -17,20 +16,17 @@ def self.run
...
-    @properties = []
   end
 
   def call
     puts "[SEEDING DATABASE]: Started - This may take several minutes."
     seed_users
     seed_campaigns
     seed_properties
-    ImpressionSeeder.run(ENV["MAX_IMPRESSIONS"].to_i)
+    ImpressionSeeder.run(ENV["MAX_IMPRESSIONS"].to_i) if ENV["MAX_IMPRESSIONS"].present?
```
brascoder: Will there be no default amount of impressions? They won't get impressions unless they specify a max?

db/seeds.rb
```diff
@@ -17,20 +16,17 @@ def self.run
   end
 
   def initialize
-    @user_id = User.last&.id.to_i + 1
-    @property_id = Property.last&.id.to_i + 1
-    @user_count = 0
+    @user_id = User.maximum(:id).to_i
```
brascoder: 👍 I didn't know `.maximum` was a thing

db/seeds.rb
hopsoft: Still debating this one

PR #62, opened 2018-11-27
Campaign show page

coderberry:
codefund_ads

codefund_ads

codefund_ads

codefund_ads

codefund_ads

codefund_ads

codefund_ads


app/controllers/campaign_budgets_controller.rb
```diff
@@ -0,0 +1,14 @@
+class CampaignBudgetsController < ApplicationController
+  before_action :authenticate_user!
+  before_action :set_campaign
+
+  def show
+  end
```
hopsoft: I've changed my thinking on having empty actions in controllers. While it does communicate some intent, the routes file is the true canonical source of what actions exist. I see these empty method defs as unnecessary noise... at least for experienced Rails developers. This practice could lead to cruft that we forget to clean up if/when the routes change.

No need to remove right now. I just wanted to communicate my thoughts on this pattern.

app/controllers/campaign_changelogs_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def create
+    comment = Comment.build_from(@campaign, current_user.id, params[:note])
+    comment.save
+    redirect_to campaign_campaign_notes_path(@campaign)
+  end
+
+  def diff
+    render layout: false
+  end
```
hopsoft: ⚠️ We should adhere to a strict REST pattern with __all__ controllers/actions. While this may seem pedantic, adhering to this strict policy will pay __big__ dividends in the future.

Consider changing to:

```ruby
# config/routes.rb
resource :diffs, only: [:show]
```

This controller/action now becomes polymorphic in nature and can show diffs on any model with a `paper_trail`.

app/controllers/campaign_changelogs_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def rollback
+    campaign = @version.reify
+    campaign.save
+    redirect_to campaign_campaign_changelogs_path(campaign)
+  end
```
hopsoft: ⚠️ We should adhere to a strict REST pattern with __all__ controllers/actions. While this may seem pedantic, adhering to this strict policy will pay __big__ dividends in the future.

Consider changing to:

```ruby
# config/routes.rb
resource :rollbacks, only: [:create]
```

This controller/action now becomes polymorphic in nature and can support rollbacks for any model with a `paper_trail`.

app/controllers/campaign_notes_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+    redirect_to campaign_campaign_notes_path(@campaign)
+  end
+
+  private
+
+  # Use callbacks to share common setup or constraints between actions.
+  def set_campaign
+    @campaign = Campaign.find(params[:campaign_id])
+  end
+end
```
hopsoft: I think we could likely have a polymorphic `comments` controller and a single set of views that work for any model we give `acts_as_commentable_with_threading` behavior to.

```ruby
# config/routes.rb
resources :campaigns do
  resources :comments
end
```

app/models/campaign.rb
```diff
@@ -134,6 +152,38 @@ def date_range=(value)
...
+    return 0 unless start_date.present? && end_date.present?
+    (end_date - start_date).to_i
+  end
+
+  def total_spend
+    1295.13
+  end
+
+  def remaining_days
+    return 0 unless start_date.present? && end_date.present?
```
hopsoft: Could reuse `total_days` impl here. 

```ruby
return 0 if total_days.zero?
```

app/models/comment.rb
```diff
@@ -0,0 +1,48 @@
...
+  scope :find_comments_for_commentable, lambda { |commentable_str, commentable_id|
+    where(commentable_type: commentable_str.to_s, commentable_id: commentable_id).order("created_at DESC")
+  }
+
+  # Helper class method to look up a commentable object
+  # given the commentable class name and id
+  def self.find_commentable(commentable_str, commentable_id)
+    commentable_str.constantize.find(commentable_id)
+  end
+end
```
hopsoft: Missing our standard model comments that encourage consistent structure.

app/models/concerns/campaigns/presentable.rb
```diff
@@ -13,5 +13,21 @@ def duration
       return nil unless start_date && end_date
       distance_of_time_in_words(start_date, end_date, scope: "datetime.distance_in_words.short")
     end
+
+    def percentage_complete_by_date
+      return 0   unless start_date.present? && end_date.present?
+      return 0   if start_date > Date.current
+      return 100 if end_date < Date.current
+
+      ((remaining_days.to_f / total_days.to_f) * 100).to_i
```
hopsoft: Probably want an extra guard to avoid divide by zero.

```ruby
return 0 if total_days.zero?
```

app/models/concerns/campaigns/presentable.rb
```diff
@@ -13,5 +13,21 @@ def duration
...
+    end
+
+    def percentage_complete_by_budget
+      return 0   unless start_date.present? && end_date.present?
+      return 0   if start_date > Date.current
+      return 100 if end_date < Date.current
+
+      ((remaining_days.to_f / total_days.to_f) * 100).to_i
```
hopsoft: See comment above about divide by zero. Note the first guard could/should be replaced with the `total_days` impl.

app/models/impression.rb
```diff
@@ -37,6 +37,7 @@ class Impression < ApplicationRecord
 
   # callbacks .................................................................
   # scopes ....................................................................
+  scope :clicked, -> { where.not(clicked_at: nil) }
```
hopsoft: We don't current have an index to optimize this query. How often will we ask for only clicks vs all impressions then determining clicks from that set? Is this something we should optimize for?

app/views/campaign_changelogs/_table.html.erb
```diff
@@ -0,0 +1,31 @@
...
+      <tr>
+        <td><%= version.index %></td>
+        <td><%= version.created_at %></td>
+        <td><%= link_to find_version_author(version), user_path(find_version_author(version)) %></td>
+        <td><%= version.event.humanize %></td>
+        <td class="text-right">
+          <a data-controller="remote-modal"
+						data-action="click->remote-modal#displayModal"
+            data-url="<%= diff_campaign_campaign_changelog_path(@campaign, version) %>"
+						href="#">
```
hopsoft: Weird formatting.

config/initializers/paper_trail.rb
```diff
@@ -0,0 +1,2 @@
+PaperTrail.config.enabled = true
+PaperTrail.config.version_limit = 3
```
hopsoft: Is 3 enough for `version_limit`? I thought we wanted to target 25.

config/routes.rb
```diff
@@ -18,7 +18,19 @@
...
+  resources :campaigns do
+    resource :campaign_targeting, only: [:show], path: "/targeting"
+    resource :campaign_budget, only: [:show], path: "/budget"
+    resources :campaign_notes, only: [:index, :create, :destroy], path: "/notes"
+    resources :campaign_properties, only: [:index], path: "/properties"
+    resources :campaign_changelogs, only: [:index, :destroy], path: "/changelog" do
+      member do
+        get :diff, to: "campaign_changelogs#diff"
+        patch :rollback, to: "campaign_changelogs#rollback"
+      end
```
hopsoft: ⚠️  Non-REST violation. See other comments about this.

app/controllers/campaign_budgets_controller.rb
coderberry: Can you help me understand this better? How would you handle this?

app/models/impression.rb
coderberry: probably

app/controllers/campaign_budgets_controller.rb
hopsoft: I'd just omit the empty method. Rails will simply render view template if the route is defined.

config/routes.rb
```diff
@@ -18,12 +18,30 @@
...
 
-  resources :campaigns
+  resources :versions, only: [:show, :update]
+  resources :comments, only: [:create, :destroy]
+
+  resources :campaigns do
+    resource :campaign_targeting, only: [:show], path: "/targeting"
+    resource :campaign_budget, only: [:show], path: "/budget"
+    resources :campaign_properties, only: [:index], path: "/properties"
+  end
```
hopsoft: These routes are beautiful.

config/routes.rb
```diff
@@ -18,12 +18,30 @@
...
   resources :impressions
+
   resources :properties do
     resources :property_screenshots, only: [:update]
   end
+  scope "/properties/:property_id" do
+    resources :versions, only: [:index], as: :property_versions
+    resources :comments, only: [:index], as: :property_comments
+  end
+
```
hopsoft: Love it.

app/models/campaign.rb
```diff
@@ -134,6 +152,38 @@ def date_range=(value)
     self.end_date   = Date.strptime(dates[1], "%m/%d/%Y")
   end
 
+  def properties
+    PropertySearch.new({
+      keywords: keywords,
+      negative_keywords: negative_keywords,
+      status: ["active"],
+    }).apply(Property.order(name: :asc))
+  end
```
hopsoft: I might rename this method to `matching_properties`. What do you think?

app/models/campaign.rb
```diff
@@ -134,6 +152,38 @@ def date_range=(value)
...
+
+  def remaining_budget
+    1599.95
+  end
+
+  def average_ctr
+    0.35
+  end
+
```
hopsoft: These methods can be replaced with my branch when it lands.

app/models/comment.rb
```diff
@@ -0,0 +1,79 @@
...
+  # to all commentable types for a given user.
+  scope :find_comments_by_user, lambda { |user|
+    where(user_id: user.id).order("created_at DESC")
+  }
+
+  # Helper class method to look up all comments for
+  # commentable class name and commentable id.
+  scope :find_comments_for_commentable, lambda { |commentable_str, commentable_id|
+    where(commentable_type: commentable_str.to_s, commentable_id: commentable_id).order("created_at DESC")
+  }
```
hopsoft: Love these scopes. I'd probably remove the prefix of `find_comments_` as its somewhat redundant and I think the semantics will read better. 

Consider this usage:

```ruby
Comment.by_user(user)
# ~vs~
Comment.find_comments_by_user(user)
```

Also, switch to stabby lambda `->` for scopes.

app/models/property.rb
```diff
@@ -75,6 +76,19 @@ class Property < ApplicationRecord
...
+    status
+    name
+    url
+    ad_template
+    ad_theme
+    language
+    keywords
+    prohibited_advertisers
+    prohibit_fallback_campaigns
+  ]
```
hopsoft: I'm might consider sorting these types of lists. Sorting makes the semantics worse, but it also helps speed up scanning for a value and spotting dupes in the list.

config/webpack/environment.js
```diff
@@ -5,26 +5,26 @@ const webpack = require('webpack');
...
-    Chartist: "Chartist",
-    Typed: "Typed",
-    SVGInjector: "SVGInjector",
-    Noty: "Noty",
+    jQuery: 'jquery',
+    'window.jQuery': 'jquery',
+    Popper: 'popper.js',
+    Chartist: 'Chartist',
+    Typed: 'Typed',
+    SVGInjector: 'SVGInjector',
```
hopsoft: Are we using `SVGInjector` or was that experimental?

app/controllers/campaign_budgets_controller.rb
brascoder: In light of the fact that this is open source, do you think that it may be confusing to less experienced Rails developers?

app/models/comment.rb
coderberry: These are gem-generated files

coderberry: I'll update it

PR #67, opened 2018-11-28
User show page enhancements

coderberry:
#67


app/controllers/application_controller.rb
```diff
@@ -5,10 +5,17 @@ class ApplicationController < ActionController::Base
...
   before_action :set_paper_trail_whodunnit
 
   protected
 
+  def authenticate_administrator!
+    unless AuthorizedUser.new(current_user).can_admin_system?
+      render status: :forbidden, text: "You do not have access to this page."
+    end
+  end
+
```
hopsoft: Note that we have an "official" forbidden page. It's currently styled like other Rails error pages, but I'm expecting that we'll customize such pages to match the site. 

```ruby
return render_forbidden
```

Should we also verify that the user isn't masquerading in our admin auth?

```ruby
return render_forbidden if user_masquerade?
return render_forbidden unless AuthorizedUser.new(current_user).can_admin_system?
```

hopsoft: Are you aware of the __pretender__ gem? https://github.com/ankane/pretender

I'm not sure how it stacks up against __devise_authorizer__ but it may be worth considering also? More stars, the docs are better, and they've addressed concerns like ActionCable.

app/controllers/users_controller.rb
```diff
@@ -58,7 +59,7 @@ def set_user_search
 
   # Use callbacks to share common setup or constraints between actions.
   def set_user
-    @user = if params[:id] == "me"
+    @user = if params[:id] == "me" || !authorized_user.can_admin_system?
```
hopsoft: This reads a little odd. I'd consider flipping the logic.

```ruby
@user = authorized_user.can_admin_system?
  User.find(params[:id])
else
  current_user
end
```

app/controllers/identicons_controller.rb
```diff
@@ -0,0 +1,7 @@
+class IdenticonsController < ApplicationController
+  def show
+    user = User.find(params[:user_id])
+    base64_identicon = RubyIdenticon.create_base64(user.email, border_size: 50, grid_size: 7, square_size: 50, background_color: 0xfafafaff)
+    send_data Base64.decode64(base64_identicon), type: "image/png", disposition: :inline
+  end
+end
```
hopsoft: We should probably cache these images.

app/javascript/src/app/controllers/fallback_image_controller.js
```diff
@@ -0,0 +1,10 @@
+import { Controller } from 'stimulus';
+
+export default class extends Controller {
+  connect() {
+    const $target = jQuery(this.element);
```
hopsoft: Lets avoid jQuery whenever possible.

```javascript
const img = event.target;
img.addEventListener('error', () => img.setAttribute('src', img.dataset.fallbackImageUrl));
```

app/javascript/src/app/turbolinks-persist-scroll.js
```diff
@@ -0,0 +1,24 @@
...
+    const elements = document.querySelectorAll(`[data-${persistScrollDataAttribute}="true"]`);
+    for (let i = 0; i < elements.length; i++) {
+      elements[i].addEventListener('click', () => {
+        enabled = true;
+      });
+    }
+
+    if (scrollPosition) window.scrollTo(0, scrollPosition);
+  });
+};
```
hopsoft: I wonder if this would work applied on `turbolinks:before-render` by setting on the new body. Just a thought. Don't worry about updating.

app/models/comment.rb
```diff
@@ -48,8 +48,8 @@ class Comment < ActiveRecord::Base
 
   # class methods .............................................................
   class << self
-    alias_method :by_user, :find_comments_by_user
-    alias_method :for_commentable, :find_comments_for_commentable
+    alias by_user find_comments_by_user
+    alias for_commentable find_comments_for_commentable
```
hopsoft: Do these need to be aliases or could the original scopes be renamed?

app/controllers/identicons_controller.rb
coderberry: How do we do this?

app/models/comment.rb
coderberry: These scopes were brought in with the gem. I'd prefer not to change in case they have internal gem calls

app/controllers/identicons_controller.rb
hopsoft: Something like this.

```ruby
cache_key = "#{user.cache_key}/identicon"
base64_identicon = Rails.cache.fetch(cache_key, expires_in: 2.weeks) do
  RubyIdenticon.create_base64(user.email, border_size: 50, grid_size: 7, square_size: 50, background_color: 0xfafafaff)
end

PR #81, opened 2018-12-06
Add ip geocoding

hopsoft:


app/controllers/advertisements_controller.rb
```diff
@@ -21,10 +21,21 @@ def set_virtual_impression_id
...
 
+  def ip_info
+    @ip_info ||= MMDB.lookup(request.remote_ip)
+  end
+
   def set_campaign
-    # TODO: smarter campaign selection
-    @campaign = Campaign.active.available_on(Date.current).for_property_id(params[:property_id]).limit(10).to_a.sample
-    @campaign ||= Campaign.active.available_on(Date.current).fallback_for_property_id(params[:property_id]).limit(10).to_a.sample
+    country_code = ip_info.country.iso_code
```
coderberry: will `MMDB.lookup` always return results even if it's the IP is bad or can't be found?

app/lib/mmdb.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    private
+
+    def mmdb_path
+      Dir.glob(Rails.root.join("db/maxmind/**/GeoLite2-City.mmdb")).first
+    end
+
+    def mmdb
+      @mmdb ||= create_mmdb(mmdb_path)
+      @mmdb = create_mmdb(mmdb_path) if refresh?
```
coderberry: smart

app/controllers/advertisements_controller.rb
hopsoft: Yes but the result may be an "empty" result but it will behave just like any other result.

PR #95, opened 2018-12-10
Add dashboards for Advertiser, Publisher and Admin

### LINKED ISSUE #93: Admin Dashboard

coderberry:
Fixes #93


app/controllers/dashboards_controller.rb
```diff
@@ -4,6 +4,9 @@ class DashboardsController < ApplicationController
 
   def show
     @context = params[:id]
+
+    @active_advertisers = User.advertisers.with_active_campaigns.order(company_name: :asc)
+    @active_campaigns = Campaign.active.order(name: :asc)
```
hopsoft: Should we also restrict campaigns to ones currently operating? Its possible to be `active` but not valid for the current day.

coderberry: I don't think so. I think `active` is good enough in this case

PR #97, opened 2018-12-11
Property receivable (amount owed)

hopsoft:


app/models/concerns/impressionable.rb
```diff
@@ -80,21 +80,35 @@ def total_impressions_per_mille
...
 
-  def daily_impressions_per_mille
-    daily_impressions_count / 1_000.to_f
+  def daily_impressions_per_mille(date = nil)
+    daily_impressions_count(date) / 1_000.to_f
   end
 
-  def dates_with_impressions
-    impressions.select(:displayed_at_date).distinct.pluck(:displayed_at_date)
+  def probable_dates_with_impressions(start_date = nil, end_date = nil)
```
coderberry: Would you mind adding a small descriptor to this method?

db/migrate/20181211201904_add_property_revenue_percentage.rb
```diff
@@ -0,0 +1,5 @@
+class AddPropertyRevenuePercentage < ActiveRecord::Migration[5.2]
+  def change
+    add_column :properties, :revenue_percentage, :decimal, null: false, default: 0.75
```
coderberry: If this is the revenue rate for a property, we should default it to 0.5

PR #110, opened 2018-12-13
Fix template format

brascoder:


coderberry:
@brascoder I'm going to make this change in another PR.

PR #111, opened 2018-12-13
Add show page for creatives displaying all templates

coderberry:
Screenshot:

image


app/controllers/creative_previews_controller.rb
```diff
@@ -0,0 +1,25 @@
+class CreativePreviewsController < ApplicationController
+  before_action :authenticate_user!
+
+  include AdRenderable
```
hopsoft: I'd move AdRenderable above the `before_action` to be the first line inside the class def.

app/controllers/concerns/ad_renderable.rb
```diff
@@ -0,0 +1,41 @@
+module AdRenderable
+  extend ActiveSupport::Concern
+
+  included do
+    attr_reader :template_name, :theme_name
+  end
```
hopsoft: These `attr_readers` may work declared in the module without needing to declare inside the `included` method.

app/controllers/creative_previews_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def show
+    @creative = current_user.creatives.find(params[:creative_id])
+    @campaign = Campaign.new(creative: @creative)
+    @template_name = params[:template] if ENUMS::AD_TEMPLATES.value?(params[:template])
+    @theme_name = params[:theme] if ENUMS::AD_THEMES.value?(params[:theme])
+
+    render_not_found unless @template_name && @theme_name
```
hopsoft: You likely want `return render_not_found`?

app/views/ad_templates/bottom-bar/show.html.erb
```diff
@@ -6,7 +6,7 @@
         <span><%= @campaign.creative.body %></span>
       </a>
       <a href="https://codefund.io" class="cf-powered-by" target="_blank" rel="noopener">
-        ads via codefund.io
+        ethical ad by CodeFund
```
hopsoft: I like this change.

app/views/creatives/show.html.erb
```diff
@@ -16,6 +16,53 @@
...
+              <div class="card-body p-2">
+                <iframe src="<%= creative_previews_path(@creative, template: template, theme: theme) %>"
+                  height="270" style="height: 320px; width: 100%; border: 0; resize: horizontal;"
+                  frameBorder="0" scrolling="no"></iframe>
+              </div>
+            </div>
+          </div>
+        <% end %>
+      <% end %>
+    </div>
```
hopsoft: Might be worth a partial here?

test/controllers/creative_previews_controller_test.rb
```diff
@@ -0,0 +1,8 @@
+require "test_helper"
+
+class CreativePreviewsControllerTest < ActionDispatch::IntegrationTest
+  test "should get show" do
+    get creative_previews_show_url
```
hopsoft: These route helpers with `show` `index` etc... always fail. Not sure why the generator is creating them like this? We just need to omit the `show` portion of the method name.

PR #113, opened 2018-12-13
Add conditional cacheing on ad preview

coderberry:


app/controllers/creative_previews_controller.rb
```diff
@@ -6,8 +6,8 @@ class CreativePreviewsController < ApplicationController
   def show
     @creative = current_user.creatives.find(params[:creative_id])
     @campaign = Campaign.new(creative: @creative)
-    @template_name = params[:template] if ENUMS::AD_TEMPLATES.values.include?(params[:template])
-    @theme_name = params[:theme] if ENUMS::AD_THEMES.values.include?(params[:theme])
+    @template_name = params[:template] if ENUMS::AD_TEMPLATES.values.to_a.include?(params[:template])
+    @theme_name = params[:theme] if ENUMS::AD_THEMES.values.to_a.include?(params[:theme])
```
hopsoft: Curious that values isn't already an Array?

coderberry: It's a fix to make standardize stop auto-fixing it incorrectly

PR #116, opened 2018-12-14
UX cleanup, add date pickers

coderberry:


coderberry:
Agreed. I'll add this in a follow-up PR

PR #117, opened 2018-12-14
Manage Applicants / Updates to Campaigns

coderberry:


app/controllers/applicant_results_controller.rb
```diff
@@ -0,0 +1,10 @@
+class ApplicantResultsController < ApplicationController
+  before_action :authenticate_administrator!
+  before_action :set_applicant
+
+  private
+
+  def set_applicant
+    @applicant = Applicant.find(params[:applicant_id])
+  end
+end
```
hopsoft: Can this be handled by `ApplicantsController`? If not I wonder if we should treat this like the other search controllers?

app/models/applicant.rb
```diff
@@ -0,0 +1,115 @@
...
+    #   *First Name:* #{advertiser_params[:first_name]}
+    #   *Last Name:*  #{advertiser_params[:last_name]}
+    #   *Email:*  #{advertiser_params[:email]}
+    #   *Company:*  #{advertiser_params[:company_name]}
+    #   *Monthly Budget:*  #{advertiser_params[:monthly_budget]}
+    #   *Website:*  #{advertiser_params[:company_url]}
+    # MESSAGE
+    # ApplicantMailer.with(form: advertiser_params.to_h).advertiser_application_email.deliver_later
+    # redirect_to advertisers_path, notice: "Your request was sent successfully. We will be in touch."
+  end
```
hopsoft: Let's name the stubs a little better.

app/controllers/applicant_results_controller.rb
coderberry: This isn't a search controller, it's a tabbed controller (like `campaign_budgets_controller.rb`)

app/models/applicant.rb
coderberry: Doh!

PR #159, opened 2018-12-21
Setup jobs for tracking daily spend

hopsoft:


app/jobs/create_debit_for_campaign_and_date_job.rb
```diff
@@ -0,0 +1,39 @@
...
+    debit = OrganizationTransaction.find_by(attrs)
+    return if debit # transaction already exists
+
+    debit = OrganizationTransaction.new(attrs.merge(
+      description: "Daily Spend for Campaign [#{campaign.id}]",
+      amount: amount,
+      posted_at: Time.current,
+    ))
+
+    campaign.organization.balance = campaign.organization.balance - amount
```
coderberry: Is this idempotent? It seems like it will deduct each time it's run from the balance

hopsoft: Look above. We return early if we find that the transaction already exists.

PR #160, opened 2018-12-22
[WIP] Enhance Org view

coderberry:


db/migrate/20181222164913_add_gift_credit_to_organization_transactions.rb
```diff
@@ -0,0 +1,6 @@
+class AddGiftCreditToOrganizationTransactions < ActiveRecord::Migration[5.2]
+  def change
+    add_column :organization_transactions, :gift_credit, :boolean, default: false
+    add_index :organization_transactions, :gift_credit
```
hopsoft: I might consider naming this column simply `gift`. This would support `gift` debits also. Probably a slim chance of this happening, but an advertiser could "gift" some money from their balance to CodeFund to distribute to open source maintainers.

app/controllers/organization_transactions_controller.rb
```diff
@@ -12,7 +12,7 @@ def index
   end
 
   def new
-    @organization_transaction = @organization.organization_transactions.build
+    @organization_transaction = @organization.organization_transactions.build(posted_at: Date.current)
```
hopsoft: Shouldn't `posted_at` be managed entirely server side when the transaction is saved? If true, this would apply to the whitelisted strong params too.

coderberry: No. Historical payments can be applied. This is especially important as we migrate data from v1.

PR #178, opened 2018-12-29
Add docsify template, legacy support and login link for mobile

coderberry:


app/controllers/advertisements_controller.rb
```diff
@@ -72,6 +72,8 @@ def prohibited_hour?
 
   # TODO: deprecate legacy support on 2019-04-01
   def property_id
+    params[:legacy_property_id] = params[:property_id] if params[:property_id] =~ /[a-f0-9]{8}-[a-f0-9]{4}-4[a-f0-9]{3}-[89aAbB][a-f0-9]{3}-[a-f0-9]{12}/
+
```
hopsoft: This line will fail for some of the routes that we have pointing at this controller. I'd update to

```ruby
params[:legacy_property_id] ||= params[:property_id] if params[:property_id].to_s =~ /.../
```

PR #196, opened 2019-01-04
Add abd tracking and custom embed events

coderberry:


app/views/advertisements/show.js.erb
```diff
@@ -1,19 +1,37 @@
...
+    if ( typeof window.CustomEvent === "function" ) return false;
+    function CustomEvent ( event, params ) {
+      params = params || { bubbles: false, cancelable: false, detail: undefined };
+      var evt = document.createEvent( 'CustomEvent' );
+      evt.initCustomEvent( event, params.bubbles, params.cancelable, params.detail );
+      return evt;
+    }
+    CustomEvent.prototype = window.Event.prototype;
+    window.CustomEvent = CustomEvent;
+  }
```
hopsoft: The polyfill adds some unnecessary bloat. I'd opt for simply creating the event like so.

```javascript
var evt = new Event("codefund");
evt.detail = { status: 'ok' };
window.dispatchEvent(evt);
```

https://developer.mozilla.org/en-US/docs/Web/API/Event/Event

db/migrate/20190104162627_add_abp_to_impressions.rb
```diff
@@ -0,0 +1,6 @@
+class AddAbpToImpressions < ActiveRecord::Migration[5.2]
+  def change
+    add_column :impressions, :abp, :boolean, default: false
+    add_index :impressions, :abp
+  end
+end
```
hopsoft: `abp` is somewhat abstract. I'd prefer we give this a more semantically meaningful name.

PR #201, opened 2019-01-07
Update README.md

max-si-m:
Fix command for copy env file


coderberry:
Thank you @max-si-m !

PR #209, opened 2019-01-09
Add statsd reporting

coderberry:


Gemfile
```diff
@@ -27,6 +27,7 @@ gem "mini_magick",                        "~> 4.9.2" # intended for use with Act
 gem "monetize",                           "~> 1.9.0"
 gem "money",                              "~> 6.13.1"
 gem "money-rails",                        "~> 1.13.0"
+gem "nsa",                                "~> 0.2.4", github: "coderberry/nsa"
```
hopsoft: What's the change requiring using a fork? Do you have a PR up so we can get back on the official GEM?

coderberry: The old gem has been abandoned. There has been a PR for a while with this change.

PR #216, opened 2019-01-11
Track uplift ad impressions

coderberry:


app/jobs/create_impression_job.rb
```diff
@@ -1,7 +1,13 @@
...
   queue_as :impression
 
-  def perform(id, campaign_id, property_id, ad_template, ad_theme, ip_address, user_agent, displayed_at_string)
+  def perform(*args)
+    # Protection for impressions during deploy. Can remove after successful deploy
+    args = args.insert(5, "false") if args.length == 8
+    create_impression(*args)
+  end
+
+  def create_impression(id, campaign_id, property_id, ad_template, ad_theme, uplift, ip_address, user_agent, displayed_at_string)
```
hopsoft: Might be safer to tack `uplift` to the end of the args and default it to `false`. Thoughts?

app/jobs/create_impression_job.rb
```diff
@@ -23,6 +29,7 @@ def perform(id, campaign_id, property_id, ad_template, ad_theme, ip_address, use
       property: property,
       ad_template: ad_template,
       ad_theme: ad_theme,
+      uplift: uplift == "true",
```
hopsoft: Should we call `uplift.to_s` here?

app/jobs/create_impression_job.rb
coderberry: oh.. good call!

app/jobs/create_impression_job.rb
coderberry: yes

PR #217, opened 2019-01-11
Jobs UI

coderberry:


db/migrate/20190103230117_create_job_postings.rb
```diff
@@ -19,6 +23,7 @@ def change
       t.string :province_code
       t.string :country_code
       t.text :url
+      t.date :posted_at_date, null: false
```
hopsoft: Is `posted_at_date` different than `created_at` and/or `start_date`?

coderberry: Yes.. If we import from external, they may have a different posted_at

PR #243, opened 2019-01-19
Setup trix editor for text areas

hopsoft:
Was yak shaving too much with tinymce so opt'd for Trix for dev speed. We can revisit later if desired.

Note: We need to add the CSS class `trix-content` whenever displaying content created with Trix.


hopsoft:
codefund ads 2019-01-19 11-36-00

app/javascript/src/app/stylesheets/components.scss
```diff
@@ -1,5 +1,5 @@
 @import './components/sparkline';
-@import url("https://cdnjs.cloudflare.com/ajax/libs/tinymce/4.9.2/skins/lightgray/skin.min.css");
+//@import url("https://cdnjs.cloudflare.com/ajax/libs/tinymce/4.9.2/skins/lightgray/skin.min.css");
```
coderberry: Remove?

hopsoft:
We can also replace toolbar icons with font-awesome or whatever we'd like.

PR #244, opened 2019-01-19
Job Wizard Preview

coderberry:


app/controllers/purchase_job_postings_controller.rb
```diff
@@ -0,0 +1,16 @@
+class PurchaseJobPostingsController < ApplicationController
```
hopsoft: I would name this with a noun given that it's a resource. Something like: `JobPostingPurchases`

```ruby
PurchaseJobPostings.create # less natural
JobPostingPurchases.create # more natural
```

PR #259, opened 2019-01-23
Cleaned up UI for Jobs (excluding job submission form)

coderberry:


app/javascript/src/app/controllers/submit_on_enter_controller.js
```diff
@@ -0,0 +1,15 @@
...
+    jQuery(this.element).keydown(function(event) {
+      const keypressed = event.keyCode || event.which;
+      if (keypressed == 13) {
+        jQuery(this.element)
+          .closest('form')
+          .submit();
+      }
+    });
+  }
+}
```
hopsoft: I'd prefer we do this without `jQuery`. Also `moment` doesn't seem to be required.

app/views/job_postings/preview.html.erb
```diff
@@ -38,7 +38,7 @@
                   <small class="h6 font-weight-light text-muted">
                     <%= link_to @job_posting.company_name, @job_posting.company_url, target: "_blank", class: "text-link" %>
                     <span class="px-1">•</span>
-                    <%= @job_posting.city %>, <%= @job_posting.country_code %>
+                    <%= @job_posting..company_location %>
```
hopsoft: Are the two dots `..` a typo?

PR #265, opened 2019-01-28
Add feature flags

coderberry:
On stage/prod, run the following:

```ruby
Flipper.register(:admins) { |actor| actor.respond_to?(:administrator?) && actor.administrator? }
Flipper.instance[:jobs].enable_group :admins
```


coderberry:
Please disregard. This is using a cannon when a rubber band gun is needed

PR #267, opened 2019-01-28
Simple feature flags

coderberry:


config/initializers/feature_flag.rb
```diff
@@ -0,0 +1,8 @@
+class FeatureFlag
+  def self.enabled?(feature, user = nil)
+    env_enabled = ENV["ENABLED_FEATURES"].split(",").map(&:to_sym)
```
hopsoft: Should we memoize this?

PR #276, opened 2019-01-30
Add metatags gem, touch-ups on job view page

coderberry:


app/controllers/application_controller.rb
```diff
@@ -9,11 +9,56 @@ class ApplicationController < ActionController::Base
...
   impersonates :user
 
   protected
 
+  # See https://github.com/kpumuk/meta-tags#allowed-options-for-display_meta_tags-and-set_meta_tags-methods
+  def set_meta_tag_data
+    set_meta_tags(
+      site: "CodeFund",
+      title: controller_name.to_s.humanize.titleize.gsub("Pages", "Home"),
+      description: "CodeFund is an open ad platform that funds developers through ethical advertising.",
```
hopsoft: Should we use the same tag line that's on the readme? Might be good to have some consistency regardless of what the tag line actually is.

> CodeFund Ads is an ethical and discreet ad platform that funds open source.

app/models/job_posting.rb
```diff
@@ -128,6 +129,15 @@ def province
     Province.find_by_iso_code(province_code)
   end
 
+  def to_meta_tags
+    {
+      title: title,
+      keywords: ["Jobs"] + keywords,
+      description: "Job Description: #{meta_description}",
+      image_src: company_logo_url,
+    }
```
hopsoft: Do we want company name here?

PR #288, opened 2019-02-04
Associate applicants with the referring campaign and property

coderberry:


app/models/applicant.rb
```diff
@@ -108,6 +111,14 @@ def user_id
     @user_id ||= (User.where(email: "eric@codefund.io").first || User.administrator.first).id
   end
 
+  def referring_property
+    @referring_property ||= Property.where(id: referring_property_id).first
+  end
```
hopsoft: `find_by(conditions)` is a bit more idiomatic than `where(conditions).first`

app/models/applicant.rb
```diff
@@ -108,6 +111,14 @@ def user_id
...
+
+  def referring_campaign
+    @referring_campaign ||= Campaign.where(id: referring_campaign_id).first
```
hopsoft: `find_by`

app/controllers/advertisers_controller.rb
```diff
@@ -5,6 +5,14 @@ def index
 
   def create
     @applicant = Applicant.new(applicant_params)
+
+    if session[:ref].present?
+      impression = Impression.where(id: session[:ref]).first
+      @applicant.referring_campaign_id = impression&.campaign_id
+      @applicant.referring_property_id = impression&.property_id
+      @applicant.referring_impression_id = impression&.id
```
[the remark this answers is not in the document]
coderberry: I suppose it doesn't hurt

PR #316, opened 2019-02-13
Initial referral program

hopsoft:


app/controllers/advertisements_controller.rb
```diff
@@ -25,7 +25,7 @@ def show
       @advertisement_html = render_advertisement
       @campaign_url = advertisement_clicks_url(@virtual_impression_id, campaign_id: @campaign.id)
       @impression_url = impression_url(@virtual_impression_id, template: template_name, theme: theme_name, format: :gif)
-      @powered_by_url = root_url(ref: @virtual_impression_id)
+      @powered_by_url = invite_url(User.referral_code(@campaign.user_id))
```
coderberry: Should this use the publisher ID and not the campaign user ID?

app/models/user.rb
```diff
@@ -212,4 +224,12 @@ def send_devise_notification(notification, *args)
   def ensure_roles
     self.roles = roles & ENUMS::USER_ROLES.values
   end
+
+  def ensure_referral_code
+    self.referral_code ||= begin
+      code = SecureRandom.urlsafe_base64(8)
```
coderberry: Is this line needed?

app/views/referrals/index.html.erb
```diff
@@ -0,0 +1,83 @@
...
+                    data-success-class="fas fa-check">
+                    <span id="linkIcon" class="fas fa-clone"></span>
+                  </a>
+                </div>
+              </div>
+            </div>
+            <small class="form-text text-muted">Copy or share your referral link with friends or followers</small>
+
+            <!-- AddToAny BEGIN -->
+            <div class="a2a_kit a2a_kit_size_32 a2a_default_style mt-3">
```
coderberry: For this to work you will need to set the meta stuff in the controller.

app/controllers/advertisements_controller.rb
hopsoft: 👍 Great catch. Thanks.

app/models/user.rb
hopsoft: Yes. It looks redundant but is required to set the initial value.

app/views/referrals/index.html.erb
hopsoft: I'll get you to walk me through that again.

hopsoft: Are meta tags not already being set implicitly? https://github.com/gitcoinco/code_fund_ads/blob/master/app/controllers/application_controller.rb#L12

PR #323, opened 2019-02-15
Add login/register via Google, GitHub and LinkedIn

coderberry:
image


app/models/user.rb
```diff
@@ -68,6 +68,7 @@ class User < ApplicationRecord
   include Users::Presentable
   include Users::Stripeable
   include Eventable
+  include FullNameSplitter
```
hopsoft: This may conflict with methods `models/concerns/users/presentable.rb`. We should probably just get rid of any similar methods in the concern.

coderberry: I don't think it conflicts actually. Only the full_name= is being introduced with the gem

PR #337, opened 2019-02-19
Advertisers can manage their campaign targeting

coderberry:


app/controllers/campaigns_controller.rb
```diff
@@ -28,7 +28,8 @@ def new
     @campaign = current_user.campaigns.build(
       status: "pending",
       start_date: Date.tomorrow,
-      end_date: 30.days.from_now
+      end_date: 30.days.from_now,
+      ecpm: ENV.fetch("BASE_ECPM", 4).to_f
```
hopsoft: Shouldn't `ecpm` be a money type?

coderberry: Yes.. good find

PR #338, opened 2019-02-20
Add links to metabase dashboards as admin

coderberry:


app/controllers/concerns/sortable.rb
```diff
@@ -11,7 +11,7 @@ def sortable_columns
   end
 
   def order_by
-    "#{sort_column} #{sort_direction}"
+    "#{sort_column} #{sort_direction} NULLS LAST"
```
hopsoft: This looks like a possible SQL injection vulnerability. We should probably wrap this with some defensive checks.

app/views/advertiser_dashboards/show.html.erb
```diff
@@ -14,6 +14,10 @@
 
 <main id="content" role="main" class="bg-light">
   <div class="container space-1">
+    
+    <% if authorized_user.can_admin_system? || current_user != true_user %>
+      <small class="text-center d-block pb-3"><%= link_to "open metabase dashboard in new window", "https://metabase.codefund.io/dashboard/10?date_filter=thismonth&advertiser_user_id=#{current_user.id}", class: "text-link", target: "_blank" %></small>
+    <% end %>
```
hopsoft: Is `current_user != true_user` too permissive here?

app/views/publisher_dashboards/show.html.erb
```diff
@@ -14,6 +14,10 @@
 
 <main id="content" role="main" class="bg-light">
   <div class="container space-1">
+    
+    <% if authorized_user.can_admin_system? || current_user != true_user %>
```
hopsoft: Same authorization related question as above.

app/views/advertiser_dashboards/show.html.erb
coderberry: Only admins can masquerade right? Is there a better way?

app/controllers/concerns/sortable.rb
coderberry: The sort column and direction are being restricted to certain fields:

```ruby
  def order_by
    "#{sort_column} #{sort_direction} NULLS LAST"
  end

  def sort_column
    return params[:column] if sortable_columns.include?(params[:column])
    return "name"          if sortable_columns.include?("name")
    "created_at"
  end

  def sort_direction
    %w[asc desc].include?(params[:direction]) ? params[:direction] : "asc"
  end
```

hopsoft: Nice job. I should have done more research.

app/views/advertiser_dashboards/show.html.erb
hopsoft: Yes. This might be a better solution. 

```
AuthorizedUser.new(true_user).can_admin_system?
```

PR #388, opened 2019-03-09
HubSpot integration

coderberry:


app/views/layouts/application.html.erb
```diff
@@ -32,7 +32,7 @@
 
     <%= noty_flash %>
     <%= content_for :render_async %>
-    <%= hubspot_tag %>
-    <%= ga_tag %>
+    <%= hubspot_tag unless authorized_user.can_admin_system? %>
+    <%= ga_tag unless authorized_user.can_admin_system? %>
```
hopsoft: Do we want/need to collect 3rd party metrics for site usage?

coderberry: For now, yes. We do not do any tracking of our ad viewers, which is why we are ethical. We also do not use generic data to do any remarketing or profiling. It's simply to help us grow as a business.

PR #390, opened 2019-03-11
Add daily summaries for impressions

### LINKED ISSUE #382: Create daily rollup tables for advertisers, publishers, campaigns, and properties
Roll data up into summary tables to alleviate pressure on the impressions table(s) and to improve overall performance.

hopsoft:
Resolves #382


app/jobs/create_daily_summary_job.rb
```diff
@@ -0,0 +1,22 @@
...
+    return if impressionable.daily_summaries.on(date).scoped_by(scoped_by).exists?
+
+    impressions = impressionable.impressions.on(date).scoped_by(scoped_by)
+
+    impressionable.daily_summaries.create!(
+      displayed_at_date: date,
+      scoped_by_type: scoped_by&.class&.name,
+      scoped_by_id: scoped_by&.id,
+      impressions_count: impressions.count,
+      clicks_count: impressions.clicked.count,
```
coderberry: What are your thoughts on adding  CTR and CPC to this as well? It seems like a simple add and it would reduce us from having to calculate it on the fly. Also, perhaps we can add Paid Ad %?

app/jobs/create_daily_summary_job.rb
```diff
@@ -0,0 +1,22 @@
...
+      estimated_gross_revenue: Money.new(impressions.sum(:estimated_gross_revenue_fractional_cents).round, "USD"),
+      estimated_property_revenue: Money.new(impressions.sum(:estimated_property_revenue_fractional_cents).round, "USD"),
```
coderberry: I wonder if we need to continue using `estimated` now that these numbers are set in stone.

app/models/concerns/impressionable.rb
```diff
@@ -1,79 +1,115 @@
...
+        .pluck(:displayed_at_date, :impressions_count)
+        .each_with_object({}) { |row, memo| memo[row[0]] = row[1] }
+      dates = (start_date..end_date).to_a
+      missing_dates = dates - counts_by_date.keys
+      if missing_dates.present?
+        CreateDailySummariesJob.perform_later self, start_date.iso8601, end_date.iso8601, scoped_by
+        impressions.on(*missing_dates).scoped_by(scoped_by).group(:displayed_at_date).count
+          .each { |date, count| counts_by_date[date] ||= count }
+      end
+      dates.map { |date| counts_by_date[date] ||= 0 }
```
coderberry: Why are we excluding 0's from this response? I'm just curious.

app/models/concerns/impressionable.rb
```diff
@@ -1,79 +1,115 @@
...
-    Rails.cache.fetch(key) { impressions.clicked.count }
+  def total_clicks_count(start_date = nil, end_date = nil, scoped_by: nil)
+    daily_clicks_counts(start_date, end_date, scoped_by: scoped_by).sum
   end
 
-  def daily_clicks_count(date = nil)
-    date = Date.coerce(date)
-    key = "#{cache_key}/#{__method__}/#{date.cache_key minutes_cached: 15}"
-    Rails.cache.fetch(key) { impressions.on(date).clicked.count }.to_i
+  def total_click_rate(start_date = nil, end_date = nil, scoped_by: nil)
```
coderberry: Should this be `average_click_rate`?

app/models/concerns/impressionable.rb
```diff
@@ -1,79 +1,115 @@
...
+    clicks_count = total_clicks_count(start_date, end_date, scoped_by: scoped_by)
+    impressions_count.zero? ? 0 : (clicks_count / impressions_count.to_f) * 100
   end
 
-  def daily_clicks_counts(start_date = nil, end_date = nil)
-    key = "#{cache_key}/#{__method__}/#{Date.coerce(start_date).cache_key}-#{Date.coerce(end_date).cache_key}"
-    Rails.cache.fetch(key) {
-      results = impressions.clicked.between(start_date, end_date).group(:displayed_at_date).count
-      (start_date..end_date).map { |date| results[date] || 0 }
+  def estimated_gross_revenue(start_date, end_date = nil, scoped_by: nil)
```
coderberry: Do we still want to use `estimated`?

app/models/concerns/impressionable.rb
```diff
@@ -1,79 +1,115 @@
...
+    start_date = Date.coerce(start_date)
+    end_date = Date.coerce(end_date || start_date)
+    key = "#{cache_key}/#{__method__}/#{start_date.cache_key}-#{end_date.cache_key}/#{scoped_by&.cache_key}"
```
coderberry: How critical is cacheing this now that we are pulling the data from roll-up tables?

app/models/concerns/impressionable.rb
```diff
@@ -1,79 +1,115 @@
...
-    (clicks_count / impressions_count.to_f) * 100
-  end
-
-  def daily_click_rates(start_date = nil, end_date = nil)
-    icounts = daily_impressions_counts(start_date, end_date)
-    ccounts = daily_clicks_counts(start_date, end_date)
-    icounts.map.with_index do |icount, i|
-      icount > 0 ? (ccounts[i] / icount.to_f) * 100 : 0
-    end
+  def estimated_property_revenue(start_date, end_date = nil, scoped_by: nil)
```
coderberry: Same question re: `estimated`

app/models/daily_summary.rb
```diff
@@ -0,0 +1,81 @@
...
+  monetize :estimated_property_revenue_cents, numericality: {greater_than_or_equal_to: 0}
+  monetize :estimated_house_revenue_cents, numericality: {greater_than_or_equal_to: 0}
+
+  # class methods .............................................................
+  class << self
+  end
+
+  # public instance methods ...................................................
+
+  def impressions_per_mille
```
coderberry: I'm not sure this name or method makes sense in terms of advertising. Is this method necessary? It's just caused me confusion a couple of times now.

app/jobs/create_daily_summary_job.rb
hopsoft: We already have `click_rate` and `ecpm`. I like the idea of adding `cpc` will get that in.

app/jobs/create_daily_summary_job.rb
hopsoft: We can remove the prefix but I'd rather leave that change for another PR as it will involve db migrations also.

app/models/concerns/impressionable.rb
hopsoft: Not sure I follow you on this question.

app/models/concerns/impressionable.rb
hopsoft: This has been renamed to simply `click_rate`

app/models/concerns/impressionable.rb
hopsoft: The daily rollup tables help a lot, but I still think it's important to cache those daily rollup calculations.

app/models/daily_summary.rb
hopsoft: It's necessary to calculate eCPM but I'll move it to be an internal implementation detail of `set_ecpm`.

hopsoft:
I'm planning on adding tests in a follow up PR, but am happy to add them to this one. I just wasn't sure if we could tolerate the delay given that this PR resolves several production issues.

app/models/concerns/impressionable.rb
hopsoft: Note: `click_rate` is not the average of daily click rates. Rather it's calculated based on impression and click counts for the given date range.

coderberry:
I agree the urgency is important, but we will need to ensure that this fix is not going to cause additional issues. If you are confident, I am as well.

Were you able to test all the pages for publishers, advertisers, and administrators with Chrome?

PR #440, opened 2019-03-24
XML job importers

brascoder:
Add functionality to download, parse, and import jobs from XML job feeds


app/jobs/import_talroo_jobs_job.rb
```diff
@@ -0,0 +1,7 @@
+class ImportTalrooJobsJob < ApplicationJob
+  queue_as :default
+
+  def perfrom(*tags)
```
coderberry: Typo

app/jobs/import_talroo_jobs_job.rb
```diff
@@ -0,0 +1,7 @@
+class ImportTalrooJobsJob < ApplicationJob
+  queue_as :default
```
coderberry: I wonder if we should set up a different queue for these, or at least set it to low

app/lib/job_feed_importers/talroo_job_feed_importer.rb
```diff
@@ -0,0 +1,143 @@
...
+          job_buffer = ''
+        end
+      end
+      req.run
+    end
+
+    private
+
+    def purge_jobs
+      JobPosting.talroo.delete_all
```
coderberry: we have links to jobs.. if we completely delete them all, we'll need to make sure we use the remote unique ID in the URL or something like that.

For example, if we import a job and the URL is '/jobs/123', then we reimport the same job the next day, we want the URL to be the same for the same job.

If this adds scope, please let me know.

app/jobs/import_talroo_jobs_job.rb
hopsoft: I think we setup a new queue with lower priority than `low`

app/lib/job_feed_importers/talroo_job_feed_importer.rb
hopsoft: Easy to accomplish this by defining `to_param` on the `JobPosting` model that has smarts about remotely sourced jobs.

app/lib/job_feed_importers/talroo_job_feed_importer.rb
```diff
@@ -34,8 +34,9 @@ def import
         end
       end
       req.run
-      ids = JobPosting.talroo.pluck(:id)
-      EnsureJobPostingsFullTextSearchJob.perform_later(ids)
+      JobPosting.talroo.pluck(:id).each_slice(1000) do |slice|
+        EnsureJobPostingsFullTextSearchJob.perform_later(slice)
+      end
```
hopsoft: Not sure `each_slice` is needed since we're using `pluck` which loads all the ids into an array.

lib/tasks/schedule.rake
```diff
@@ -16,20 +16,52 @@ namespace :schedule do
...
+      EnsureDailySummariesJob.perform_later
+    end
+
+    desc <<~DESC
+      Import jobs from Talroo
+      NOTE: Schedule daily
+    DESC
+    task import_talroo: :environment do
+      tags = ENUMS::KEYWORDS.values.flatten.uniq.sort
+      ImportTalrooJobsJob.new.perform(*tags)
```
hopsoft: We need to update to `perform_later`

lib/tasks/schedule.rake
```diff
@@ -16,20 +16,52 @@ namespace :schedule do
...
+      puts "There are #{JobPosting.count} jobs"
+    end
+
+    desc <<~DESC
+      Import jobs from ZipRecruiter
+      NOTE: Schedule daily
+    DESC
+    task import_zip_recruiter: :environment do
+      tags = ENUMS::KEYWORDS.values.flatten.uniq.sort
+      ImportZipRecruiterJobsJob.new.perform(*tags)
```
hopsoft: We need to update to `perform_later`

app/jobs/create_talroo_job_posting_job.rb
```diff
@@ -0,0 +1,82 @@
...
+  SOURCE = ENUMS::JOB_SOURCES::TALROO
+
+  def perform(xml, tags)
+    setup_tags(tags)
+
+    fragment = Nokogiri::XML(xml)
+    return unless fragment.children.present?
+
+    job = extract_job_params(fragment)
+    JobPosting.create(job) unless job.nil?
```
hopsoft: We may want to call `create!` here... though it could be noisy if there are lots of errors on save. I'm just worried that we may swallow records that don't save due to a smaller validation error or something. Thoughts?

PR #476, opened 2019-04-19
Add additional filters, sorting defaults, fix bug with applicant filter

coderberry:


app/controllers/campaign_searches_controller.rb
```diff
@@ -33,6 +34,9 @@ def campaign_search_params
       negative_keywords: [],
       province_codes: [],
       statuses: [],
-    )
+    ).tap do |whitelisted|
+      whitelisted[:fallback] = false if whitelisted[:fallback] == "No"
+      whitelisted[:fallback] = true if whitelisted[:fallback] == "Yes"
+    end
```
hopsoft: Could simplify to

```ruby
    ).tap do |whitelisted|
        whitelisted[:fallback] = whitelisted[:fallback] == "Yes"
     end
```

coderberry: I don't think that will work. We need to be able to set it to true, false or not set it at all

PR #482, opened 2019-04-29
Add budget tests in context of ad rendering

hopsoft:
Cleanup for faster ad rendering and to reduce Rollbar noise
and cognitive load.


test/integration/advertisements_test.rb
```diff
@@ -324,4 +324,33 @@ class AdvertisementsTest < ActionDispatch::IntegrationTest
     assert response.body.include?("template=text&theme=light")
     assert response.body =~ /house: true/
   end
+
+  # ----------------------------------------------------------------------------------------------------------
+
+  test "js: premium ads don't render when budgets available" do
```
coderberry: Probably remove `don't`?

PR #494, opened 2019-05-17
Use Hubspot form in place of internal applicant form

coderberry:


app/views/layouts/application.html.erb
```diff
@@ -15,13 +15,18 @@
     <%= stylesheet_pack_tag 'code_fund', media: 'all', 'data-turbolinks-track': 'reload' %>
 
     <!-- Compressed JavaScripts -->
+    <script
+      src="https://code.jquery.com/jquery-3.4.1.min.js"
+      integrity="sha256-CSXorXvZcTkaix6Yvo6HppcZGetbYMGWSFlBw8HfCJo="
+      crossorigin="anonymous"></script>
```
hopsoft: I'm pretty certain we can reuse the jQuery already included with Bootstrap rather than loading it twice. Something like this perhaps.

```js
// app/javascript/src/app/index.js
import jQuery from 'jquery';
window.jQuery = window.$ = jQuery;
```

Better would be to leverage Webpack foo for this. Something along these lines. https://stackoverflow.com/questions/29080148/expose-jquery-to-real-window-object-with-webpack#29083197

Issue #402, opened 2019-03-14
Convert country targeting to tree-based checkboxes


coderberry:
Replace the existing country selection with a tree-based version.

As an admin,
* if I select a subregion, it selects all countries in the subregion
* if I select a subregion, I can deselect countries within the subregion

![image](https://user-images.githubusercontent.com/12481/54373269-913d2480-4642-11e9-9b09-16ebda456735.png)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 2.14 ETH (500.08 USD @ $233.68/ETH)  attached to it as part of the gitcoinco fund.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/code_fund_ads/402/3003).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/gitcoinco/code_fund_ads/402/3003).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $83,568.97 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 1 week, 6 days from now.
Please review their action plans below:


**1) [brascoder](https://gitcoin.co/profile/brascoder) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/gitcoinco/code_fund_ads/402/3003?mutate_worker_action=approve&worker=brascoder) | [reject worker](https://gitcoin.co/issue/gitcoinco/code_fund_ads/402/3003?mutate_worker_action=reject&worker=brascoder))_.**

Update markup to use a checkbox tree
Add stimulus controllers to manage the tree and the branches

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/code_fund_ads/402/3003).

gitcoinbot:
Issue Status: 1. Open 2. **Cancelled** 

<hr>

__The funding of 2.14 ETH  (519.7 USD @ $242.85/ETH) attached to this issue has been **cancelled** by the bounty submitter__

 
[...]

PR #501, opened 2019-05-23
Change country selection to checkbox tree

brascoder:
Addresses #402 

<a href="https://cl.ly/0c6dbe90c03a" target="_blank"></a>

### DIFF
--- a/app/controllers/campaigns_controller.rb
+++ b/app/controllers/campaigns_controller.rb
@@ -48,6 +48,7 @@ class CampaignsController < ApplicationController
 
   def create
     @campaign = Campaign.new(campaign_params)
+    @campaign.country_codes = @campaign.country_codes.filter { |code| code.present? }
 
     respond_to do |format|
       if @campaign.save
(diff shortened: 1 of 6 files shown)


app/controllers/campaigns_controller.rb
```diff
@@ -48,6 +48,7 @@ def edit
 
   def create
     @campaign = Campaign.new(campaign_params)
+    @campaign.country_codes = @campaign.country_codes.filter { |code| code.present? }
```
hopsoft: This is great but consider `@campaign.country_codes.select(&:present?)` to be a bit more idiomatic.

app/javascript/src/app/controllers/checkbox_tree_branch_controller.js
```diff
@@ -0,0 +1,41 @@
...
+      return this.leafTargets.every((leaf) => {
+        return leaf.checked;
+      });
+    };
+
+    const anyChecked = () => {
+      return this.leafTargets.some((leaf) => {
+        return leaf.checked;
+      });
+    }
```
hopsoft: I might consider moving these helper methods out as first class _private_ methods of the controller.

PR #504, opened 2019-05-23
Add unique_ip_addresses_count to daily_summaries

hopsoft:
Also obfuscate stored ip addresses


app/jobs/create_impression_job.rb
```diff
@@ -27,7 +31,7 @@ def perform(id, campaign_id, property_id, ad_template, ad_theme, ip_address, use
       property: property,
       ad_template: ad_template,
       ad_theme: ad_theme,
-      ip_address: ip_address,
+      ip_address: Digest::MD5.hexdigest("#{ip_address}#{ip_address_salt}"),
```
coderberry: Will this break our call for Maxmind data?

hopsoft: No this happens after all the Maxmind work has been performed.

PR #510, opened 2019-05-29
Distinct mobile ads

hopsoft:


app/controllers/advertisements_controller.rb
```diff
@@ -180,8 +181,9 @@ def keywords
   end
 
   def set_campaign
-    return nil if request.bot?
+    return nil if device.bot?
     return nil unless property.active? || property.pending?
+    return nil if device_small? && property.hide_on_responsive?
```
coderberry: 💙

app/views/advertisements/show.js.erb
```diff
@@ -40,13 +40,23 @@
       xhr.send();
     }
 
+    function closeCodefundAd() {
```
coderberry: It's nitpicky, but I wonder if we should use `closeCodeFundAd()`?

PR #519, opened 2019-06-05
Restore ability for advertisers to manage creatives & images 

coderberry:


app/models/creative.rb
```diff
@@ -51,6 +53,14 @@ class << self
 
   # public instance methods ...................................................
 
+  def pending?
+    status == ENUMS::CREATIVE_STATUSES::PENDING
+  end
+
+  def locked?
+    !pending?
+  end
```
hopsoft: Shouldn’t this just be the inverse of active? i.e. archived is considered locked too?

PR #520, opened 2019-06-05
Add support to exclude properties in a campaign

coderberry:


coderberry:
Verbal approval from Nate

PR #532, opened 2019-06-13
Replace Disqus with Commento and Command+K for search

coderberry:


package.json
```diff
@@ -18,6 +18,7 @@
     "filestack-js": "^1.10.0",
     "git": "^0.1.5",
     "github-buttons": "git+https://github.com/ntkme/github-buttons.git",
+    "hotkeys-js": "^3.6.11",
```
hopsoft: Not sure a new dependency is required for this.

```js
document.addEventListener('keydown', (e) => {
  if (e.metaKey && e.code === 'KeyK') { 
    console.log('Command + K pressed'); 
  }
});
```

app/javascript/src/app/controllers/search_autocomplete_controller.js
```diff
@@ -37,6 +39,19 @@ export default class extends Controller {
...
 
+  initCommandK() {
+    hotkeys('command+k', function(event, handler) {
+      // Prevent the default refresh event under WINDOWS system
+      event.preventDefault();
+      jQuery('#btn-search-toggle')
+        .data('HSUnfold')
+        .show();
+      setTimeout(function() {
+        jQuery('#search-input').focus();
```
hopsoft: Probably better to use explicit Stimulus targets rather than `id`s to access elements.

If the desired elements aren't part of the Stimulus controller, I generally wrap them in getter methods that match the Stimulus targets interface.

```js

get someElementTarget() {
  return document.getElementById('some-element');
}
```

package.json
coderberry: I tried that and it didn't work

hopsoft: Hmm... I wrote that and tested in a browser without issue. 

Maybe we have another js dependency that's cancelling the default event behavior. My guess is the lib you added must be opting to capture the event and handle it before passing off to other handlers.

app/javascript/src/app/controllers/search_autocomplete_controller.js
hopsoft: See the `useCapture` argument here: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

package.json
hopsoft: See the `useCapture` argument here: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

coderberry: I got it working :D

Issue #483, opened 2019-05-01
Allow forcing fallback ads during local development


mglaman:
## User Story

As a _publisher_, I want to force using fallback ads locally so that I do not hit real ads during local development.

## Description

Since CodeFund has fallback ads, I was wondering if there was a way to "force" CodeFund to server a fallback ad. This would be useful when developing locally.

## Acceptance Critera

Although I cannot find official AdSense docs, there appears to be `var google_adtest = "on"` for sync and `data-adtest="on"` for async. 

What if we could pass a query param:

```
https://codefund.app/properties/{id}/funder.js?adtest=true
```

## Related

List any URL's or other issues that provide more context.

hopsoft:
This is a great idea and will be simple to implement. Also note that properties should not show premium ads until they have been set to `active`.

mglaman:
Would it be possible to check "This is not the property domain, only show fallback ads"? That way it is just seamless. I guess this is hard as properties only have `URL` and not a domains property.

Maybe step 1: flag. step 2: check if a valid domain.

hopsoft:
All good ideas. Thank you for brainstorming on this with us.

mglaman:
🙌 thanks for the amazing service.

I'd contribute a PR, but I don't know a lick of Ruby.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.417 ETH (99.9 USD @ $239.57/ETH)  attached to it as part of the gitcoinco fund.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/gitcoinco/code_fund_ads/483/3103).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/gitcoinco/code_fund_ads/483/3103).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $78,335.21 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 4 weeks, 1 day from now.
Please review their action plans below:


[...]

PR #544, opened 2019-06-18
Allow forcing fallback ads

mglaman:
Allows `adtest` param to force fallback ad campaigns.

PR for https://github.com/gitcoinco/code_fund_ads/issues/483.


app/controllers/advertisements_controller.rb
```diff
@@ -197,9 +201,14 @@ def set_campaign
       .targeted_country_code(country_code)
       .targeted_province_code(province_code)
 
-    @campaign = get_premium_campaign(geo_targeted_campaign_relation) if property.active?
-    @campaign ||= get_fallback_campaign(geo_targeted_campaign_relation)
-    @campaign ||= get_fallback_campaign(campaign_relation)
+    if !adtest.nil?
```
mglaman: Breaks if the value was `false.`

app/controllers/advertisements_controller.rb
```diff
@@ -185,6 +185,10 @@ def referral_code
     @referral_code ||= User.referral_code(property.user_id)
   end
 
+  def adtest
+    @adtest ||= params[:adtest]
```
mglaman: Need this to be a truthy value, always string symbol.

mglaman:
@coderberry yay! 🙌 first time working with Ruby beyond configuring Compass way back. Thanks for the merge!

PR #549, opened 2019-06-20
Take more control over multivariate tests

hopsoft:
Should get conversion tracking of alternatives working in production.
Has the added benefit of not using andy form of user persistence.


app/controllers/application_controller.rb
```diff
@@ -16,6 +16,9 @@ class ApplicationController < ActionController::Base
 
   protected
 
+  def track_multivariate_conversion(impression_id)
```
coderberry: Is this intentionally empty? May be add a comment if so?

hopsoft: woops... its an artifact. will remove.

PR #561, opened 2019-06-28
[WIP] Fix campaign targets

brascoder:


coderberry:
Sorry @brascoder but I had to fix this myself. We need the fixes live today for the launch into July.

Issue #391, opened 2019-03-13
Property not properly displaying on events screen


JasonBarnabe:
As a publisher, the events views I see on URLs like https://codefund.app/properties/XXXXX/events does not properly show my property beside each event. Instead I see entries like this:

```
Jason Barnabe
#<Property:0x00007f8876b2b960>
Property was created
about 1 hour ago
```

coderberry:
@JasonBarnabe This is great feedback. We will add it to our list of items to do. As you know we are also open source. If you want to submit a PR with this change that would be great!

andrewmcodes:
@hopsoft this seems to be working correctly now and should be safe to close.

PR #577, opened 2019-07-25
Display property name in events

brascoder:
Address #391 

This change displays the name of the property on each event. That might be a little redundant, so if there is something else you would like displayed, let me know


hopsoft:
Let's move the `to_s` implementation to `Properties::Presentable` and make it an alias of `analytics_key`. Thanks.

PR #579, opened 2019-07-26
Build in-app reports for campaigns

hopsoft:


app/helpers/application_helper.rb
```diff
@@ -1,6 +1,12 @@
 module ApplicationHelper
   include Pagy::Frontend
 
+  def country_display_name(iso_code)
+    country = Country.find(iso_code)
+    return "#{country.emoji_flag} #{truncate country.name, length: 30} (#{iso_code})" if country
```
coderberry: Thank you!

app/models/daily_summary_report.rb
```diff
@@ -0,0 +1,118 @@
...
+
+  # scopes ....................................................................
+
+  default_scope -> {
+    paid_impressions_count = Arel::Nodes::Subtraction.new(
+      arel_table[:impressions_count].sum,
+      arel_table[:fallbacks_count].sum
+    )
+
+    select(:impressionable_type, :impressionable_id, :scoped_by_type, :scoped_by_id)
```
coderberry: This is why you make the big bucks ;)

PR #581, opened 2019-07-30
Campaign in-app creative reports

hopsoft:


app/controllers/campaign_creatives_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+  before_action :set_campaign
+
+  def index
+    @summary_report = @campaign.summary_report(@start_date, @end_date)
+    @reports = @campaign.creative_reports(@start_date, @end_date)
+
+    respond_to do |format|
+      format.html
+      format.csv do
+        send_data(
```
coderberry: can you please set the filename on this to include campaign id and date range?

Example: `campaign-country-report-231-201907010-20190731.csv`

This will help a lot when we send these reports to clients, and will allow them to keep them organized

Issue #587, opened 2019-08-06
Yesterday's earnings stats are set to zero


idoshamun:
## User Story

As a publisher, I want to see yesterday's earning stats.

## Description

Every day, yesterday's earning stats are broken until noon (Israel time). Then they go back to normal and show the real information.
This is very odd because if I look at today's stats they look perfect and when today becomes yesterday they are set to zero.

## Acceptance Critera

Yesterday's stats should be ok during the whole day.

hopsoft:
I'll take a stab at explaining what's happening technically. _This isn't a solution only a description of what's causing the observed behavior._

- We store all data in UTC and reports are also based on UTC
- We have a background task that rolls up daily data that kicks off at 9:00 AM UTC
- Our reporting dashboard shows **historical data** from our rollups OLAP and **current day** activity from our "transactions" OLTP

hopsoft:
This is also causing overspend for high traffic sites. /cc @coderberry

andrewmcodes:
@hopsoft is this still an issue?

idoshamun:
I think that it is already fixed :)

andrewmcodes:
Awesome!

PR #594, opened 2019-08-22
Restrict overspend to a 10 minute window max...

hopsoft:
This should resolve high campaign overspend in cases where there is no
daily budget on a campaign that is showing on high traffic sites
(especially those with large traffic spikes).


hopsoft:
This will mitigate the overspend caused by the combo of no/high daily caps + #587

Issue #574, opened 2019-07-23
"Open" should be "Open Source"


coderberry:
Here's what we have currently:

![image](https://user-images.githubusercontent.com/12481/61734556-6032d600-ad3f-11e9-9d75-5eb6a8cbd9e8.png)

This should say "Open Source" and not "Open" in both the header and content.

PR #605, opened 2019-09-10
Update bottom-bar z-index

coderberry:


EduardoGHdez:
I guess this solves #574

Issue #567, opened 2019-07-10
Formatting does not exist for email confirmation page


coderberry:
https://codefund.io/users/confirmation?confirmation_token=XXX

PR #606, opened 2019-09-10
[FIX] Add styling to reset password confirmation page

swanny85:
Proposed fix for #567. I used the `devise/passwords/new.html.erb` as a template.
confirmation
I also replaced the devise links as those were also pointing to Oauth I saw you all discontinued.  I replaced it with a login link.

### DIFF
--- a/app/views/devise/confirmations/new.html.erb
+++ b/app/views/devise/confirmations/new.html.erb
@@ -1,20 +1,37 @@
-<h2>Resend confirmation instructions</h2>
+<!-- ========== MAIN ========== -->
+<main class="content" role="main" class="border-bottom">
+  <div class="container space-top-t space-top-md-4 space-bottom-2">
+      <%= simple_form_for(resource, as: resource_name, url: confirmation_path(resource_name), html: { method: :post, class: "js-validate w-md-75 w-lg-50 mx-md-auto" }) do |f| %>
+        <!-- Title -->
+        <div class="mb-7">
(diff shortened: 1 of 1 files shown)


scswan85:
#567

dahal:
@scswan85 including screenshot will prolly make review process much easier. I would include one?

It prolly will help contributes if there was a proper template when submitting  a PR @codeberry @hopsoft

app/views/devise/confirmations/new.html.erb
```diff
@@ -1,20 +1,37 @@
...
-                autofocus: true,
-                value: (resource.pending_reconfirmation? ? resource.unconfirmed_email : resource.email),
-                input_html: { autocomplete: "email" } %>
+        <!-- Button -->
+      <div class="row align-items-center mb-5">
+        <div class="col-3 col-sm-4">
+          <%= link_to "Login", new_user_session_path, class: "small link-muted" %>
+        </div>
+        <div class="col-9 col-sm-8 text-right">
+          <button type="submit" class="btn btn-secondary transition-3d-hover">Resend Confirmation Instructions</button>
```
dahal: Instead of hardcoding these `Resend Confirmation Instructions` also to keep it dry, I think you should use something like `I18.t('devise.confirmation.instruction')` or  `t('devise.confirmation.instruction')` (for short)

andrewmcodes:
This has been fixed in #648.

PR #610, opened 2019-09-11
Allow admins to reinvite users within profile page

### LINKED ISSUE #607: Add "reinvite" button to user profile (admin view only)
## User Story

As an administrator, I can view a user's profile and click the button 'Re-invite'. That will trigger the call `user.invite!`. Then I will see a flash message stating they have been reinvited.

dahal:
Closes #607

Re-invite users

### DIFF
--- a/app/controllers/administrator_invitations_controller.rb
+++ b/app/controllers/administrator_invitations_controller.rb
@@ -1,0 +1,17 @@
+class AdministratorInvitationsController < ApplicationController
+  before_action :authenticate_administrator!
+  before_action :set_user
+
+  # PUT /admin_invitations/:id
+  # PATCH /admin_invitations/:id
+  def update
+    @user.invite!
+    flash[:success] = t('devise.invitations.send_instructions', email: @user.email)
+    redirect_to user_path(@user)
+  end
+
+  private
+  def set_user
+    @user = User.find(params[:id])
+  end
+end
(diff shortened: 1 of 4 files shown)


hopsoft:
Awesome. Will review this tomorrow.

app/controllers/administrator_invitations_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+  # PATCH /admin_invitations/:id
+  def update
+    @user.invite!
+    flash[:success] = t('devise.invitations.send_instructions', email: @user.email)
+    redirect_to user_path(@user)
+  end
+
+  private
+  def set_user
+    @user = User.find(params[:id])
```
dahal: In case you are wondering why I didn't user find_by, I want this to throw `ActiveRecord::RecordNotFound` exception if this user is not found, we should prolly have a proper/global (controller/concern) `ExceptionHandler` (if you don't have one already) so we don't have to check/handle on each controller/action.

config/routes.rb
```diff
@@ -40,6 +40,7 @@
   resource :pricing, only: [:show]
   resource :search, only: [:show], controller: :search
   resources :websites, only: [:index]
+  resources :administrator_invitations, only: [:update], path: "/administrator/invitations"
   resource :administrator_dashboards, only: [:show], path: "/dashboards/administrator"
```
dahal: I am just going with the flow here, but I was kinda curious, why not namespace these controllers, I mean something like:-

```
namespace : administrator do
end

coderberry:
This is a beautiful UX implementation. Well done @dahal!

dahal:
@hopsoft @coderberry i was wondering if there is anything that needs to be done on this PR or is this good to go?

PR #611, opened 2019-09-11
Block property inline on property report (admin only)

coderberry:


app/javascript/src/app/controllers/inline_toggle_controller.js
```diff
@@ -8,33 +8,18 @@ export default class extends Controller {
...
       method: 'PATCH',
       dataType: 'json',
       credentials: 'same-origin',
       headers: { 'X-CSRF_Token': Rails.csrfToken(), 'Content-Type': 'application/json' },
-      body: JSON.stringify({
-        job_posting: {
-          auto_renew: checked,
-        },
-      }),
+      body: JSON.stringify({ checked: checked }),
```
hopsoft: Could be shortened to 

```javascript
body: JSON.stringify({ checked }),
```

app/models/campaign.rb
```diff
@@ -242,6 +242,16 @@ def prohibited_properties
...
 
+  def prohibit_property!(property_id)
+    ids = (prohibited_property_ids.compact << property_id.to_i).uniq.sort.compact
+    update(prohibited_property_ids: ids)
+  end
+
+  def permit_property!(property_id)
+    ids = (prohibited_property_ids.compact - [property_id.to_i]).uniq.sort.compact
+    update(prohibited_property_ids: ids)
+  end
```
hopsoft: Might consider `update_columns` here to avoid validation and callbacks.

PR #615, opened 2019-09-15
[FIX] Aws::S3::Errors::InvalidAccessKeyId when trying to run seeds

andrewmcodes:
The commit fixes the Aws::S3::Errors::InvalidAccessKeyId error when trying to run seeds in development. Since not every user will have correct AWS credentials, we need to be using disk instead of S3 for ActiveStorage in development. 

Since some users may need to use S3 in development, this also adds some enhancements to the example file workflow. Users can now use a binstub to copy example files for them. This will be important when adding a docker workflow since a lot of these config files will be different whether you are developing in docker or on your local machine and it will be important to allow config files to be set up to allow users to do one or the other.

Ex: 

```sh
➜ ./bin/copy_example_files

== Copying example files ==
config/storage.yml
.env

== Example files copied ==
```

This PR, in conjunction with #590 should allow users to run `bin/rails db:seed` without error 🎉


hopsoft:
@andrewmcodes I think we can close this PR in favor of #633.

PR #616, opened 2019-09-15
[WIP] dockerize the app

andrewmcodes:
# TODO

- [x] resolve bundler 2.0.2 error (can fix it but isn't consistent - this actually brings in a good question of should we be bundling in containers vs image. Tradeoffs for both - happy to discuss more)
- [x] resolve redis config error
- [x] add documentation
- [x] figure out best solution for .env file
- [ ] increase `max_locks_per_second`
- [x] system tests
- [ ] remove some files

## Recommended Tools

[lazy docker](https://github.com/jesseduffield/lazydocker)


andrewmcodes:
# [WIP] Documentation 

## ENV file

You need the following in your .env file

```
COMPOSE_HTTP_TIMEOUT=300
DEPENDENCY_TIMEOUT=300s
NODE_ENV=development
PG_PASSWORD=password
PG_USERNAME=postgres
PG_HOST=postgres
RAILS_ENV=development
# note these two are already defined in this file so comment the other instances out
# or replace 127.0.0.1 with redis
REDIS_CACHE_URL=redis://redis:6379  
REDIS_QUEUE_URL=redis://redis:6379
```

## Steps
1. `bin/copy_example_files`
2. make sure .env file is updated 
3. run `docker-compose build` - note: this will take a few minutes
4. when the build is done, run `docker-compose up -d`
5. If you are starting for the first time, once your containers are up and running, run `docker-compose exec app bash`
6. In this bash session, run `bin/rails db:create db:migrate`
7. Seed if needed or just create a dump of your local db and use postgres to dump the data into the postgres container
8. Go to localhost:3000 and you should be up and running

## Gotchas
1. You always need to be inside of the app to run commands like `bin/rails db:migrate etc.`
2. You need to always prefix rails commands with `bin/`. This may be able to be fixed, need to look into it more.
3. It seems there are sometimes issues with Bundler 2.0.2, haven't quite figured this out yet.
4. Access to the file system is slow but that is normal, we can speed up with something like [blitz](https://github.com/codetales/blitz)

coderberry:
@andrewmcodes I'd love to see this one make it over the finish line if you get a chance 🙌

andrewmcodes:
@coderberry just working on system tests then figure out the best solution for the bundle issue and it should be good to review 👍

PR #617, opened 2019-09-17
Generate Campaign Report PDF

coderberry:
The code in this PR needs some cleanup, but it's for an immediate need in generating reports for our advertisers.


app/controllers/application_controller.rb
```diff
@@ -137,4 +138,20 @@ def sample_requests_for_scout
...
+      # Variable hash
+      default_url_options = {host: url}
+
+      # Overwriting current variables
+      Rails.application.config.action_controller.asset_host = url
+      Rails.application.config.action_mailer.asset_host = url
+      Rails.application.routes.default_url_options = default_url_options
+      Rails.application.config.action_mailer.default_url_options = default_url_options
+    end
+  end
```
hopsoft: I wonder if this would be better in an initializer. Would possibly require a restart if/when devs enable ngrok, but I think that would be ok.

app/jobs/generate_organization_report_job.rb
```diff
@@ -0,0 +1,30 @@
+class GenerateOrganizationReportJob < ApplicationJob
+  queue_as :default
+
+  def perform(id:, report_url:)
+    organization_report = OrganizationReport.find(id)
+    organization_report.update(status: "generating")
```
hopsoft: May want a defensive check here to ensure it's not already generated or generating.

Not sure about the available statuses but something like this would work.

```ruby
organization_report = OrganizationReport.where(status: "new").find_by(id: id)
return unless organization_report
```

app/jobs/generate_organization_report_job.rb
```diff
@@ -0,0 +1,30 @@
...
+      tempfile.binmode
+      tempfile.write(response)
+      tempfile.rewind
+
+      organization_report.pdf.attach(io: tempfile, filename: filename, content_type: "application/pdf")
+      organization_report.update(status: "ready")
+    rescue => ex
+      Rails.logger.error(ex)
+      organization_report.update(status: "error")
+    end
```
hopsoft: You'll want an ensure block that closes the tempfile.

test/fixtures/organization_reports.yml
```diff
@@ -0,0 +1,30 @@
...
+  end_date: 2019-09-16
+  campaign_ids: MyText
+  title: MyString
+
+two:
+  organization_id: 
+  start_date: 2019-09-16
+  end_date: 2019-09-16
+  campaign_ids: MyText
+  title: MyString
```
hopsoft: I'd delete the unused fixtures.

app/controllers/application_controller.rb
coderberry: I'll hold off on this till you can work your magic

PR #649, opened 2019-10-07
make pending tooltip text for properties more descriptive

### LINKED ISSUE #392: Publishers are not told what the properties being in "Pending" status means
As a publisher who just signed up, I see a yellow circle on the dashboard beside by property's name. Hovering over the yellow circle, I see the tooltip "Pending". However there is no indication what "Pending" means, or what (if anything) I need to do to stop being "Pending".

andrewmcodes:
Wasn't quite sure the best way to handle this or the proper verbiage.

There were a few ways I thought about for handling this but choose this method due to the ability to modify for other statuses in the future and since `property_status_html` was no longer being used. 

I chose the verbiage I did based on this comment in #392 from @coderberry 
>Just so you know, "Pending" means that it is awaiting approval from myself.

fixes #392


app/helpers/properties_helper.rb
```diff
@@ -23,13 +23,9 @@ def property_responsive_behaviors_for_select
...
-  def property_status_html(status)
-    case ENUMS::PROPERTY_STATUSES[status]
-    when "pending" then tag.span(class: "fas fa-circle text-warning", title: "Pending", data: tooltip_expando)
-    when "active" then tag.span(class: "fas fa-circle text-success", title: "Active", data: tooltip_expando)
-    when "rejected" then tag.span(class: "fas fa-circle text-danger", title: "Rejected", data: tooltip_expando)
-    when "archived" then tag.span(class: "fas fa-circle text-muted", title: "Archived", data: tooltip_expando)
-    when "blacklisted" then tag.span(class: "fas fa-circle text-dark", title: "Blacklisted", data: tooltip_expando)
-    end
+  def property_status_title(status)
+    return status.humanize unless status.inquiry.pending?
```
hopsoft: Nice... ActiveSupport's `inquiry` is black magic I've not used before.

PR #651, opened 2019-10-07
only display fallback ads on localhost

### LINKED ISSUE #592: Display fallback ads when viewing via localhost
As a developer when I run my app using localhost, I want to only see fallback ads, even if my property is enabled.

andrewmcodes:
fixes #592


app/controllers/advertisements_controller.rb
```diff
@@ -216,6 +216,8 @@ def referral_code
   end
 
   def ad_test?
+    return true if request.host == "localhost"
```
hopsoft: We might also want to add an ngrok check.

andrewmcodes: @hopsoft We could use `request.local?` but that caused a few tests to fail so I went with this safer approach. I think it was just due to the way the tests were set up though. Thoughts?

hopsoft: I'd prefer `request.local?` if its not too difficult to patch up the tests.

Issue #389, opened 2019-03-09
Optimizations (requested by @xhmikosr)


coderberry:
### Image Optimization

- [ ] Ensure all images are optimized using ImageOptim

#### IE10 support?

`var evt = new Event('codefund');` does not work on IE 10

#### Endpoint slowdown

This is due to CloudFlare. It's likely not something that is continuing once we stopped proxying through CloudFlare.

#### Explain 2 calls to px.js

See https://gtmetrix.com/reports/www.bootstrapcdn.com/LS2xJp7Q#waterfall

#### specify the image dimensions in the HTML. Because the HTML is drawn first, this avoids reflows

finally one last thing, cause I need to head out for a few hours: you should definitely specify a cache-control header for your files, even if it's `max-age: 0` which personally I wouldn't go with 0 if the files are cache busted or don't change like real-time

add `no-sniff` for the images too because it seems you serve them without extension

to removing the useless quote escapes, replace `0px` with `0`

XhmikosR:
I want to add another thing. We need to verify the 3rd parameter in all `xhr.open` calls is either set to `true` or absent (which means it's true IIRC)

XhmikosR:
One last thing: Any specific reason you wait for `DOMContentLoaded`?

BTW I edited my previous post with more stuff.

PR #660, opened 2019-10-09
css and ad optimizations

andrewmcodes:
partial fix for #389

See [here](https://bitfalls.com) for why the addition of `position: relative` was needed.


app/views/ad_templates/sponsored-text/show.html.erb
```diff
@@ -2,7 +2,7 @@
   <ins id="cf">
     <span class="cf-wrapper">
       Sponsored:&nbsp;
-      <%= image_tag @creative.icon_image.cloudfront_url, class: "cf-img", style: "width: 20px; height: 20px;", alt: @creative.name rescue nil %>
+      <%= image_tag @creative.icon_image.cloudfront_url, height: 20, width: 20, class: "cf-img", style: "width: 20px; height: 20px;", alt: @creative.name rescue nil %>
```
XhmikosR: When you pass the width and height attributes, you should remove the properties from `style`. IMO all images should have width and height attributes so that browsers can calculate their dimensions and avoid reflows

PR #666, opened 2019-10-10
Fix all lints flagged by eslint

andrewmcodes:
I tested everything that I could - this should all still work but please give it a thorough read.


.eslintrc.json
```diff
@@ -0,0 +1,13 @@
...
+    "es6": true,
+    "jquery": true
+  },
+  "globals": {
+    "Rails": "readonly",
+    "StripeCheckout": "readonly",
+    "Turbolinks": "readonly",
+    "Typed": "readonly"
+  }
+}
```
hopsoft: YAML please

app/javascript/src/app/controllers/animated_type_controller.js
```diff
@@ -2,12 +2,16 @@ import { Controller } from 'stimulus'
...
+    const strings = JSON.parse(this.element.dataset.strings)
 
-    new Typed(this.element, { strings, typeSpeed, loop, backSpeed, backDelay })
+    const typeSpeed = parseInt(this.element.dataset.typeSpeed)
+
+    const loop = this.element.dataset.loop === 'true'
+
+    const backSpeed = parseInt(this.element.dataset.backSpeed)
+
+    const backDelay = parseInt(this.element.dataset.backDelay)
```
hopsoft: This type of yak shaving was the basis of my Twitter hot take this morning. 🤷‍♂

app/javascript/src/app/controllers/campaign_form_controller.js
```diff
@@ -40,16 +39,12 @@ export default class extends Controller {
...
-        (html => {
-          this.setCreativeOptions(html)
-        }).bind(this)
-      )
+      .then(response => {
+        return response.text()
+      })
+      .then(html => {
+        this.setCreativeOptions(html)
+      })
```
hopsoft: Let's be sure to test this form thoroughly prior to merge.

hopsoft: Basically everything that ESLint prompted for a `bind` removal.

andrewmcodes: What if we just disable this rule if we don't feel comfortable about it?

PR #669, opened 2019-10-11
Dockerize

andrewmcodes:
Get docker set up for the app. Instructions added in the readme.


config/webpacker.yml
```diff
@@ -58,9 +58,9 @@ development:
   # Reference: https://webpack.js.org/configuration/dev-server/
   dev_server:
     https: false
-    host: localhost
+    host: <%= ENV["WEBPACKER_DEV_SERVER_HOST"] %>
     port: 3035
-    public: localhost:3035
+    public: <%= ENV["WEBPACKER_DEV_SERVER_URL"] %>
```
hopsoft: We might want to use `fetch` and return `localhost` when the environment variable isn't available.

andrewmcodes:
It looks like the webpacker host is not correctly getting interpolated into the container and the webpacker container is not being used - I will investigate.

PR #680, opened 2019-10-16
Automated Reports

coderberry:
This PR will allow advertisers or administrators to establish automated email reports.


Gemfile.lock
```diff
@@ -634,6 +634,7 @@ DEPENDENCIES
   email_address (~> 0.1.11)
   factory_bot_rails
   faker
+  foundation_emails
```
hopsoft: Wonder if we should we get rid of **premailer** or **inky** as we now have multiple gems that all seem to serve the same purpose?

test/factories/scheduled_organization_reports.rb
```diff
@@ -0,0 +1,15 @@
...
+
+    subject { "My Scheduled Report" }
+    start_date { date }
+    end_date { start_date.advance(months: 1) }
+    recipients { ["eric@codefund.io", "nate@codefund.io"] }
+    frequency { ENUMS::SCHEDULED_ORGANIZATION_REPORT_FREQUENCIES.values.sample }
+    dataset { ENUMS::SCHEDULED_ORGANIZATION_REPORT_DATASETS.values.sample }
+    campaign_ids { organization.campaigns&.pluck(:id) }
+  end
+end
```
hopsoft: Factories should be reserved for seeds only as we're striving to keep the test suite as fast as possible. I may circle back and switch this to use fixtures.

test/test_helper.rb
```diff
@@ -14,6 +15,8 @@
 # Minitest::Reporters.use! Minitest::Reporters::ProgressReporter.new
 
 class ActiveSupport::TestCase
+  include FactoryBot::Syntax::Methods
+
```
hopsoft: See note above regarding fixtures.

app/jobs/schedule_organization_report_job.rb
```diff
@@ -0,0 +1,60 @@
+class ScheduleOrganizationReportJob < ApplicationJob
+  queue_as :default
+
+  def perform(id:, ts: nil)
```
hopsoft: `ts` is a bit vague. I'd prefer to see something more descriptive.

Gemfile.lock
coderberry: *premailer* and *inky* do different things. I believe we use premailer within our ad rendering don't we? foundation_emails is the updated version of inky I believe. This can be removed though

PR #685, opened 2019-10-16
Add more smarts to sponsor campaign clicks

hopsoft:


bin/copy_production_db_to_local
```diff
@@ -1,4 +1,5 @@
 #!/bin/sh
+source .env
```
andrewmcodes: 🎉

PR #701, opened 2019-10-21
New marketing pages

coderberry:


hopsoft:
I had the same question for Eric yesterday, but this is in fact what we want. The reason is that we will be going live with the new marketing pages before we go live with the app redesign.

PR #705, opened 2019-10-22
campaigns#index redesign

andrewmcodes:
image


app/views_redesigned/campaigns/index.json.jbuilder
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+
+json.array! @campaigns, partial: "campaigns/campaign", as: :campaign
```
hopsoft: Are we missing the `app/views_redesigned/campaigns/campaign.json.jbuilder` partial?

andrewmcodes: Ah good catch, I accidentally removed it from this PR. I’ll add that in and merge.

PR #708, opened 2019-10-22
Update to Ruby 2.6.5

hopsoft:


test/test_helper.rb
```diff
@@ -17,6 +17,11 @@
 class ActiveSupport::TestCase
   include FactoryBot::Syntax::Methods
 
+  if ENV["PARALLEL_TESTS"] && (workers = Concurrent.processor_count / 2) > 1
```
andrewmcodes: We need to check if `ENV["PARALLEL_TESTS"] == 'true'` unless we just want to check that this variable exists in the environment, which is fine but I want to make sure that's what you intended. 

`if ENV["PARALLEL_TESTS"]` will evaluate to true whether the variable equals true or false.

hopsoft: I generally prefer a truthy check for the existence for env vars that are boolean like this.

PR #712, opened 2019-10-23
Properties#Show - Details tab redesign

cmckee-dev:
Properties#Show - Details

Screenshot_2019-10-24_12-02-48


app/views_redesigned/async_property_card_footers/show.html.erb
```diff
@@ -0,0 +1,51 @@
...
+      <h5 class="small font-weight-normal mb-0">
+        <span class="d-block">Current</span>
+        <span class="d-block">Campaign</span>
+      </h5>
+      <div class="text-dark mt-2"><%= @property.current_sponsor_campaign&.name || "none" %></div>
+    </div>
+
+  <% else %>
+
+    <div class="col-4 u-ver-divider">
```
andrewmcodes: `u-ver-divider` is not a class in our new theme. You could copy it in as a utility class or perhaps there is another way we could achieve this?

app/views_redesigned/async_property_card_footers/show.html.erb
```diff
@@ -0,0 +1,51 @@
...
+    </div>
+    <div class="col-4 u-ver-divider">
+      <h5 class="small font-weight-normal mb-0">
+        <span class="d-block">Average</span>
+        <span class="d-block">Click Rate</span>
+      </h5>
+      <div class="text-dark mt-2"><%= number_to_percentage @property.click_rate(@start_date, @end_date), precision: 2 %></div>
+    </div>
+    <div class="col-4">
+      <h5 class="small font-weight-normal mb-0">
```
andrewmcodes: maybe just make this a `<p>` instead of adding `small` and `font-weight-normal` here.

app/views_redesigned/properties/_standard_card.html.erb
```diff
@@ -0,0 +1,36 @@
...
+        <h6 class="mb-0">
+          <%= tag.span "", class: "fas fa-circle small mr-1 text-#{property_status_color(property.status)}",
+            title: property_status_title(property.status), data: tooltip_expando %>
+          <%= link_to property.name, property, class: "link-muted" %>
+        </h6>
+        <div class="position-relative">
+          <%= link_to property_property_screenshots_path(property), method: :put, class: "btn btn-xs btn-icon btn-light transition-3d-hover" do %>
+            <span class="fas fa-camera btn-icon__inner"></span>
+          <% end if authorized_user.can_admin_system? %>
+          <%= link_to property.url, class: "btn btn-xs btn-icon btn-light transition-3d-hover", target: "_blank" do %>
```
andrewmcodes: `transition-3d-hover` is not a class in our new theme.

app/views_redesigned/properties/_standard_card.html.erb
```diff
@@ -0,0 +1,36 @@
...
+            <span class="fas fa-external-link btn-icon__inner"></span>
+          <% end %>
+        </div>
+      </div>
+    <% end %>
+  <% end %>
+
+  <%= card_body do %>
+    <%= link_to image_tag(property.screenshot.attached? ? property.screenshot : "missing-screenshot.jpg", class: "img-fluid"), property %>
```
andrewmcodes: Let's not wrap this in `card_body` so it can be full width. See [here](https://github.com/gitcoinco/code_fund_ads/blob/0185565413cbee92aacb7d2568693e6246aa1ec8/app/views_redesigned/campaigns/_standard_card.html.erb#L10) for an example.

app/views_redesigned/async_property_card_footers/show.html.erb
andrewmcodes: Same comment to the other occurrences of this in this file.

app/views_redesigned/async_property_card_footers/show.html.erb
cmckee-dev: I actually looked at this. Its a fairly small stand alone Sass file. Maybe 30-50 lines. It looks decently clean without porting this and with Eric going through things after us, I wasn't sure how much we want to dive into UI things. Let me know your thoughts @andrewmcodes @hopsoft @coderberry

app/views_redesigned/properties/_standard_card.html.erb
cmckee-dev: Good catch!

app/views_redesigned/async_property_card_footers/show.html.erb
cmckee-dev: Same questions as above.

app/views_redesigned/properties/_standard_card.html.erb
cmckee-dev: Sounds good. Made the change.

app/views_redesigned/async_property_card_footers/show.html.erb
andrewmcodes: My opinion is we should be cleaning as much as we can (within reason). Aka trying to use consistent styling on each element and component. Along with this is creating helpers and meta classes. Ie if we use the same style on every h1, which we theoretically should, pulling those classes into our stylesheet under a new h1 class. We should be tweaking as little as possible. Hopefully that makes sense.

app/views_redesigned/async_property_card_footers/show.html.erb
andrewmcodes: I’d say first make sure our new theme doesn’t have something similar. Then if you think it is something that adds value pull the stylesheet in.

cmckee-dev: Diving into this it will take some time to port. The `u-ver-divider` style is only 50 lines but relies on other scss partials. Parent and sibling divider styles, media-breakpoints, variables, etc.

We may want to backlog this for another PR when we cleanup any vendor or dangling stylesheets.

app/views_redesigned/async_property_card_footers/show.html.erb
cmckee-dev: Diving into this it will take some time to port. The `u-ver-divider` style is only 50 lines but relies on other scss partials. Parent and sibling divider styles, media-breakpoints, variables, etc.

We may want to backlog this for another PR when we cleanup any vendor or dangling stylesheets.

app/views_redesigned/async_property_card_footers/show.html.erb
cmckee-dev: Diving into this it will take some time to port. The `u-ver-divider` style is only 50 lines but relies on other scss partials. Parent and sibling divider styles, media-breakpoints, variables, etc.

We may want to backlog this for another PR when we cleanup any vendor or dangling stylesheets.

andrewmcodes: Just remove these classes and we’ll move forward from there.

andrewmcodes:
@cmckee-dev can you updated the screenshot when you’re done please?

PR #716, opened 2019-10-23
change from render to render partial

andrewmcodes:


app/helpers/application_helper.rb
```diff
@@ -18,7 +18,7 @@ def safe_camo(url)
   def page_heading(action, subject, subtitle: nil, icon: nil)
     render("/@shared/page_heading", action: action, subject: subject, subtitle: subtitle, icon: icon) unless ENV["REDESIGN"] == "true"
 
-    render "shared/page_heading", action: action, subject: subject, subtitle: subtitle, icon: icon
+    render partial: "shared/page_heading", locals: {action: action, subject: subject, subtitle: subtitle, icon: icon}
```
hopsoft: I'd prepend the path with `/`

PR #721, opened 2019-10-23
Properties#Show - Instructions tab redesign

cmckee-dev:
Screenshot_2019-10-23_14-23-10

Depends on PR #719


app/views_redesigned/async_property_card_footers/show.html.erb
```diff
@@ -45,4 +45,3 @@
     </div>
   <% end %>
 </div>
-
```
andrewmcodes: Can you revert this change? We would like to always have a new line at the end of each file.

cmckee-dev: Oops. Fixed.

PR #727, opened 2019-10-24
campaigns#edit and campaigns#new redesign

andrewmcodes:
image
image


app/helpers/application_helper.rb
```diff
@@ -2,6 +2,8 @@ module ApplicationHelper
   include Pagy::Frontend
   include CardHelper
   include NavTabsHelper
+  include ActionsHelper
+  include ButtonHelper
```
hopsoft: These explicit `includes` shouldn't be necessary as Rails will implicitly load and include all helpers defined under `app/helpers` automatically.

PR #732, opened 2019-10-25
Property#Index redesign

cmckee-dev:
Screenshot_2019-10-24_20-02-15  

---  

Screenshot_2019-10-24_20-03-14  

---

Ran across this bug I could not track down. The search icon from the shared partial displays differently for Campaigns and Properties. While the form is collapsed the icon doesnt align correctly. Will need second pair of eyes on this. @andrewmcodes @hopsoft 

**Note:**  The spacing mis-alignment between the two screens is due to responsiveness of shrinking the window on the right.   

Screenshot_2019-10-24_19-36-13


app/views_redesigned/properties/index.html.erb
```diff
@@ -0,0 +1,16 @@
+<div class="d-flex">
+  <%= render "shared/sidebar" %>
+  <div class="container mt-4">
+    <header class="page-title-bar">
+      <%= page_heading "Viewing", "Properties",
+      subtitle: "#{pluralize(@pagy.count, "record")} found",
+      icon: "fas fa-browser" %>
+    </header>
+    <%= render partial: "/property_searches/form",
+                 layout: "/@shared/collapsable_search_form",
```
andrewmcodes:
```suggestion
                 layout: "/shared/collapsable_search_form",
```

app/views_redesigned/async_property_stat_cards/show.html.erb
```diff
@@ -0,0 +1,14 @@
+<div class="card-deck d-block d-lg-flex card-lg-gutters-3 mb-3">
+  <%= render "/@shared/widgets/stat_card",
```
andrewmcodes: you'll need to create this file and then change this throughout this file to 
```suggestion
  <%= render "/shared/widgets/stat_card",
```

app/views_redesigned/properties/_tr.html.erb
```diff
@@ -0,0 +1,65 @@
...
+      <span class="opacity-30">N/A</span>
+    <% else %>
+      <%= render_later async_property_click_rate_path(property_id: property.id) do %>
+        <i class="fa fa-spinner fa-spin text-primary"></i>
+      <% end %>
+    <% end %>
+  </td>
+
+  <td class="align-middle text-right text-secondary">
+    <%= render "/@shared/table_buttons/edit", link: edit_property_path(property) %>
```
andrewmcodes: This should be:

```suggestion
    <%= render "/shared/table_buttons/edit", link: edit_property_path(property) %>
```

but let's wait until [this](https://github.com/gitcoinco/code_fund_ads/pull/729/files#diff-ec3a97e817fce8b2ade68f6a7c5c7687R13) method is added and use it instead.

andrewmcodes: If you rebase with master, you can use this instead
```suggestion
    <%= edit_button(link: edit_property_path(property)) %>
```

app/views_redesigned/properties/index.html.erb
cmckee-dev: :heavy_check_mark:

app/views_redesigned/async_property_stat_cards/show.html.erb
cmckee-dev: :heavy_check_mark:

app/views_redesigned/properties/_tr.html.erb
cmckee-dev: :heavy_check_mark:

PR #738, opened 2019-10-25
StimulusReflex site search

hopsoft:
- Supports same style global search as before
- Supports `<CMD+K>`
- Something about the template JavaScript conflicts with the behavior of `data-reflex-permanent` as the search text gets wiped out on first search. Subsequent searches work as expected.
- Positioning is currently hard coded
- Some styling is currently inlined

CodeFund ❤️ Open Source 2019-10-25 16-44-10


andrewmcodes:
> Something about the template JavaScript conflicts with the behavior of data-reflex-permanent as the search text gets wiped out on first search. Subsequent searches work as expected.

Did you get this resolved? IMO we would probably want to get this fixed before calling this a done deal

hopsoft:
Yes. It's actually an issue with properly managing `focus`. I resolved it on the expo form validation demo and will be pulling that fix over. Context: https://github.com/hopsoft/stimulus_reflex_expo/blob/master/app/javascript/lib/focus.js

PR #770, opened 2019-10-31
add revenue percentage validation to property

andrewmcodes:


app/models/property.rb
```diff
@@ -50,11 +50,12 @@ class Property < ApplicationRecord
   # validations ...............................................................
   # validates :ad_template, presence: true
   # validates :ad_theme, presence: true
-  validates :language, length: {maximum: 255, allow_blank: false}
-  validates :name, length: {maximum: 255, allow_blank: false}
+  validates :language, length: {maximum: 255}, allow_blank: false
+  validates :name, length: {maximum: 255}, allow_blank: false
```
andrewmcodes: I changed these bc that's what they look like in the docs.

hopsoft: good catch

PR #771, opened 2019-10-31
First stab at nginx config

hopsoft:


config/nginx.conf.erb
```diff
@@ -0,0 +1,77 @@
...
+    location ~ \/impressions\/.+\/uplift {
+      root /app/public;
+    }
+
+    location ~ \/impressions\/.+\/click {
+      root /app/public;
+    }
+
+    location / {
+      proxy_pass https://codefundwww.wpengine.com;
```
coderberry: Can we make this https://wp.codefund.io ? That will send the traffic through cloudflare, but it will also give us more control and better metrics.

PR #783, opened 2019-11-05
Teams

### LINKED ISSUE #784: Implement Teams to allow groups of people to manage campaigns etc...
CodeFund ❤️ Open Source 2019-11-05 14-56-43

andrewmcodes:
## In Scope

- [x] we need the ability to switch organization/team if assigned to multiple
- [x] we need the ability to manage organization/team members and their roles

## Out of Scope

- no need for new UI to create organizations/teams. we're only working with existing orgs
- no need to create new users. we're only working with existing users
- no need to support invites with mailers
- sharing images with teammates

Resolves #784


app/controllers/comments_controller.rb
```diff
@@ -41,7 +41,7 @@ def set_organization
     @commentable = if authorized_user.can_admin_system?
       Organization.find(params[:organization_id])
     else
-      current_organization
+      Current.organization
     end
   end
```
hopsoft: I'm thinking that we'll want to extract `set_organization` out into a concern as this behavior will be needed on every controller after login.

andrewmcodes:
We need to talk about scoping things like campaigns. Should they be scoped on the user still or on org now?

app/controllers/application_controller.rb
```diff
@@ -2,6 +2,7 @@ class ApplicationController < ActionController::Base
   include Pagy::Backend
   include Authorizable
   include Dateable
+  include CurrentOrganizationable
```
hopsoft: Not in love with this concern name.

app/controllers/concerns/current_organizationable.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def find_organization
+    id = organization_id if valid_user_organization?(organization_id)
+    Organization&.find(id)
+  end
+
+  def organization_id
+    return params["current-organization"] if params["current-organization"]
+
+    if controller_name.inquiry.organizations? && params[:action] != "index"
```
hopsoft: I'd use `action_name` vs `params[:action]`

app/controllers/concerns/current_organizationable.rb
```diff
@@ -0,0 +1,55 @@
...
+      current_user&.organization
+    else
+      current_user.default_organization
+    end
+  end
+
+  def valid_user_organization?(org_id)
+    return true if authorized_user.can_admin_system?
+
+    current_user.organizations.map(&:id).include?(org_id.to_i)
```
hopsoft: I'd change to `current_user.organizations.where(id: org_id).exists?`

app/controllers/concerns/current_organizationable.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+end
```
hopsoft: Much of the logic in this concern seems like it could be condensed to improve readability with a more linear/procedural approach. Consider some of the side by side comparison critiques of the Sandi Metz OOP style of lots of little methods vs a more linear approach using local variables. I expect that if you reworked this module, you'd find that you can reduce the code by half or more while making it more understandable. Anyway, it's not a hill I'd die on right now but will simply encourage you to try the thought experiment and draw your own conclusions.

app/controllers/impersonations_controller.rb
```diff
@@ -12,4 +13,10 @@ def destroy
     stop_impersonating_user
     redirect_to users_path
   end
+
+  private
+
+  def clear_session_organization_id
+    session[:organization_id] = nil
+  end
```
hopsoft: Should we need to set this to something other than `nil` given that there should be a default organization we can use?

app/controllers/organization_users_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  private
+
+  def set_organization
+    @organization = Organization.find(params[:organization_id])
+  end
+
+  def set_organization_user
+    @organization_user = OrganizationUser.find(params[:id])
+  end
```
hopsoft: Security seems loose on these `set` methods. Should we be restricting to `current_user` or `admin`?

app/models/current.rb
```diff
@@ -1,2 +1,9 @@
 class Current < ActiveSupport::CurrentAttributes
+  attribute :organization
+
+  resets { Time.zone = nil }
+
+  def organization=(organization)
+    super
+  end
```
hopsoft: This setter seems redundant.

app/models/current.rb
```diff
@@ -1,2 +1,9 @@
 class Current < ActiveSupport::CurrentAttributes
+  attribute :organization
+
+  resets { Time.zone = nil }
```
hopsoft: Is this something we want?

app/controllers/impersonations_controller.rb
andrewmcodes: Once we clear the session, rails will redirect, hit the application controller again and set a new org if that makes sense.

hopsoft: Makes sense but when an impersonation ends, we aren't wiping the entire session are we?

andrewmcodes: No just the organization id.

app/models/current.rb
andrewmcodes: That is what I got from the docs..All we need is the organization, not any methods hanging off of it.

app/controllers/concerns/current_organizationable.rb
andrewmcodes: I added some changes...not sure if I made it better or worse. There were a lot of edge cases I encountered which is why this logic got as large as it did. I have actually refactored it several times at this point. Happy to take suggestions if you are still unhappy.

PR #798, opened 2019-11-08
New ad formats

### LINKED ISSUE #735: Develop new ad templates that use JS
## User Story

As a publisher, I want to select an ad template that will only display when the the visitor scrolls down. It drops down from the top (with a close button)

## Description

There are a handful of new templates we should explore. Here's some examples written by @octavioamu 

[codefund_templates.zip](https://github.com/gitcoinco/code_fund_ads/files/3772805/codefund_templates.zip)

hopsoft:
Resolves #735 

- [x] Restructure how ads get rendered
- [x] Translate all templates and themes over
- [x] Get ad-previews working
- [x] Get creative previews working
- [x] Update forking behavior for sponsor campaigns
- [x] Update ad API endpoints
- [x] Fix broken tests
- [x] Create floating bottom-right ad format
- [x] Create top-bar ad format
- [x] Create smart-bar ad format
- [x] Wait for ad to become visible before tracking an impression
- [x] Style refinements and tests


app/javascript/advertisements/helpers.js
```diff
@@ -0,0 +1,23 @@
...
+  }
+  return true
+}
+
+export const closeCodefund = (codefundElementId = 'codefund') => {
+  sessionStorage.setItem('codefund', 'closed')
+  document.getElementById(codefundElementId).remove()
+}
+
+export const codefundClosed = () =>
```
andrewmcodes: Pedantic and more of a question:

Wouldn’t this variable be better written as `getCodeFundClosed` and the above as `setCodeFundClosed`?

app/javascript/advertisements/index.js
```diff
@@ -0,0 +1,62 @@
+import { setUpliftUrls, detectUplift } from './uplift'
+import {
+  closeCodefund,
+  codefundClosed,
+  dispatchCodefundEvent,
+  elementVisible
+} from './helpers'
+
+const showCodeFundAdvertisement = (
```
andrewmcodes: In some places you use `Codefund` and in others `CodeFund`. Any reason?

hopsoft: Mostly artifacts from previous code. Will get everything consistent before this wraps up.

app/javascript/advertisements/helpers.js
hopsoft: Not really a setter in the pure sense, but I agree that current impl names aren’t great.

app/controllers/advertisement_previews_controller.rb
```diff
@@ -30,13 +52,28 @@ def theme_name
     ENUMS::AD_THEMES[params[:theme]]
   end
 
+  def sponsor?
+    request.format == "svg"
```
andrewmcodes: Will this always be true?

app/javascript/advertisements/media/theme.scss
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  .cf-powered-by {
+    display: block;
+    font-size: 11px;
+    margin-top: 5px;
+  }
+}
+
+// #cf[data-template='media'][data-theme='light'] {
```
andrewmcodes: Any reason this is all commented out?

app/jobs/download_and_extract_maxmind_file_job.rb
```diff
@@ -63,6 +63,6 @@ def untar(tarred_path)
         end
       end
     end
-    FileUtils.rm_rf tarred_path
+    # FileUtils.rm_rf tarred_path
```
andrewmcodes: Should this still be commented out or just removed?

config/webpack/environment.js
```diff
@@ -1,30 +1,30 @@
...
+    }
+  }
+})
 
 environment.plugins.prepend(
   'Provide',
   new webpack.ProvidePlugin({
     jQuery: 'jquery',
-    '$': 'jquery',
+    $: 'jquery',
```
andrewmcodes: Why was this changed?

app/jobs/download_and_extract_maxmind_file_job.rb
hopsoft: We can probably remove, but I think leaving it commented for now is ok. For reference, it eliminates the race condition when spinning everything up on a single machine.

app/javascript/advertisements/media/theme.scss
hopsoft: Look like @coderberry refactored this stylesheet? My guess is that this could be deleted now?

app/controllers/advertisement_previews_controller.rb
hopsoft: Perhaps not always but this impl matches the advertisements controller. I'm fine leaving it hard coded for now.

config/webpack/environment.js
hopsoft: Dunno. My guess is the JavaScript formatter did this?

hopsoft: Verified. Prettier is updating the format here.

app/javascript/advertisements/media/theme.scss
coderberry: Nope. This should have remained. Right about this point is when my wife said "You need to be done with work now". You'll likely need to uncomment and verify it looks good.

PR #822, opened 2019-11-19
Updating Authentication Redesign

cmckee-dev:
### Enhancement / Fix

This PR updates styles for authentication pages for users who aren't logged in. It removes the site header and footer to give us a middle ground between public facing and our web app styles. 

This also removes Omni-auth references from the application. CVE-2015-9284

### General

### CodeFund Core Team

## Screenshots

Every authentication page looks the same as previously minus removing the header and footer.

Screenshot_2019-11-19 Sessions


config/application.rb
```diff
@@ -27,16 +27,9 @@ class Application < Rails::Application
...
-      [Devise::SessionsController,
-       Devise::RegistrationsController,
-       Devise::ConfirmationsController,
-       Devise::UnlocksController,
-       Devise::InvitationsController,
-       Devise::PasswordsController,
-       UserPasswordsController,].each do |views|
-        views.layout proc { |controller| Rails.env.development? && ENV["REDESIGN"] == "true" ? "application_redesign" : "authentication" }
+      if Rails.env.development? && ENV["REDESIGN"]
+        ApplicationController.layout "application_redesign"
```
hopsoft: Nice.

app/views_redesigned/devise/passwords/new.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= render "devise/layout" do %>
+<%# <%= render "devise/layout" do %1> %>
```
andrewmcodes: If we aren't using this, go ahead and remove the commented out code.

app/models/user.rb
```diff
@@ -153,13 +153,11 @@ class User < ApplicationRecord
...
     :invitable,
     :lockable,
-    :omniauthable,
     :recoverable,
     :rememberable,
     :timeoutable,
     :trackable,
-    :validatable,
-    omniauth_providers: [:github]
+    :validatable
```
andrewmcodes: I believe we should have a trailing comma here

cmckee-dev: `./bin/standardize` yells at us for this.

config/application.rb
cmckee-dev: I think this is a good move for when we move the redesign to stage/prod. It allows us to easily remove the excess code.

app/views_redesigned/devise/passwords/new.html.erb
cmckee-dev: Nice :+1:

PR #824, opened 2019-11-20
Adding Global Stats Api endpoints

cmckee-dev:
### Feature

This PR adds the endpoints for querying global application stats. Stats made available are Impressions Total Count, Impressions Click Count and Impressions Click Rate. Also added and endpoint to return all stats.

### Checklist

- Created Api controller module dir

- Added GlobalStats controller file
- Added GlobalStats controller with all_global_stats method
- Added GlobalStats controller with impressions_total_count method
- Added GlobalStats controller with impressions_click_count method
- Added GlobalStats controller with impressions_click_rate method
- Added private methods to GlobalStats controller for db queries

- Added GlobalStats helper file
- Added GlobalStats test file
- Added successful response tests for each GlobalStats method

### General

### CodeFund Core Team


config/routes.rb
```diff
@@ -29,6 +29,13 @@
     omniauth_callbacks: "omniauth_callbacks",
   }
 
+  scope "/api", module: :api, defaults: {format: :json} do
+    get "/all_global_stats", to: "global_stats#all_global_stats"
+    get "/impressions_total_count", to: "global_stats#impressions_total_count"
+    get "/impressions_click_count", to: "global_stats#impressions_click_count"
+    get "/impressions_click_rate", to: "global_stats#impressions_click_rate"
+  end
+
```
hopsoft: We need to keep our endpoints restful instead of RPC style. Something more along the lines of `global_stats#show` that returns everything. 

Also, I don't think our API surface area is large enough to warrant special distinction in a subdirectory yet.
I'd move this up as a first class citizen under `controllers` rather than under the `api` directory. _Consider that the API is simply an implementation detail of which templates render after the action method completes._

test/controllers/global_stats_controller_test.rb
```diff
@@ -0,0 +1,24 @@
+require "test_helper"
+
+class Api::GlobalStatsControllerTest < ActionDispatch::IntegrationTest
+  # Test Successful Responses
```
andrewmcodes: nitpick but we don't need this comment.

app/helpers/global_stats_helper.rb
```diff
@@ -0,0 +1,2 @@
+module GlobalStatsHelper
```
andrewmcodes: Do we need this file? It seems that since this is being used as an API endpoint and there are not any views utilizing this there is no reason for this `ActionView::Helper` module.

config/routes.rb
cmckee-dev: Awesome. Refactored for the restful design like we talked about and removed the api namespace structure.

test/controllers/global_stats_controller_test.rb
cmckee-dev: Good catch, made the comments while pseudocoding the tests. Removed :+1:

app/helpers/global_stats_helper.rb
cmckee-dev: Nope. Removed the file. :+1:

app/controllers/application_controller.rb
```diff
@@ -52,6 +53,13 @@ def clear_searches(except: [])
     end
   end
 
+  def allow_cors_requests
+    if Rails.env.development?
```
andrewmcodes: Should this only be in development? Also, maybe we only need to allow it for your GlobalStatsController?

`controller_name.inquiry.global_stats?`

config/routes.rb
```diff
@@ -34,6 +34,10 @@
     invitations: "invitations",
   }
 
+  scope "/api", defaults: {format: :json} do
```
andrewmcodes: I don't think you need this scope.

app/controllers/application_controller.rb
cmckee-dev: Last week I kept running into issues in development because of localhost ports being considered CORS. I think once its in stage/prod we shouldn't run into those problems. Subdomains and nginx route filtering shouldn't run into these issues.

We can easily add this to the site as a global config. Although I think keeping CORS locked down to the minimal required for stage/prod is best. Let me know what you think.

config/routes.rb
cmckee-dev: Youre right, its not necessary.

Although think we should keep it this way. It gives clarity to the routes and url since no html pages will be served. It also allows us to define other endpoints with the same clarity and pattern in the future. Making it easier for developers to know where things go. 

Its a team decision though. I understand both sides.

andrewmcodes: Bc the controller is not in the API scope as of now (I know it was at one point), I don't think this will work. I generally would prefer it be in an api module but since @hopsoft does not I think we need to remove this scope or it won't work if that makes sense.

app/controllers/application_controller.rb
andrewmcodes: ok then lets do this. Remove this if statement and then do

`before_action :allow_cors_requests, unless: -> { Rails.env.production? }`

config/routes.rb
cmckee-dev: Yes, I did have it in a module previously. After a pair session with Nate he didnt think it was needed yet because our controllers are still manageable.  

The `scope` definition is different from `namespace` if that's what you're thinking of. Its similar to `namespace` but you define the attributes/settings. In our case nothing other than the actual route was defined. We have not defined `module: :api` as a parameter so it doesn't look for it.  

Again I'm fine with either way.

app/controllers/application_controller.rb
cmckee-dev: Nice. Clean. Will change this.

config/routes.rb
andrewmcodes: oh I see. @hopsoft want to make a ruling?

hopsoft: Let omit the scope for now as it's not used yet.

cmckee-dev: Awesome :+1: Making the changes.

config/routes.rb
```diff
@@ -34,6 +34,8 @@
     invitations: "invitations",
   }
 
+  get "/api/global_stats" => "global_stats#show", :defaults => {format: :json}
```
andrewmcodes:
```suggestion
  resource :global_stats, only: [:show]
```

PR #833, opened 2019-11-26
Implement URL blocklist

andrewmcodes:
# Enhancement

## Description

Adds the ability to block certain URL's from having ad's render on their site.

## Why should this be added

We have an issue with a domain using another properties funder link to serve ads on their site. This prevents them from doing so.

### General

### CodeFund Core Team


.standard.yml
```diff
@@ -15,3 +15,5 @@ ignore:
     - Layout/AlignHash
   - 'app/**/*':
     - Naming/AccessorMethodName
+  - 'app/controllers/advertisements_controller.rb':
+    - Performance/InefficientHashSearch
```
andrewmcodes: Had to add this bc Standard was trying to correct something and if the correction was used, an error was thrown.

config/enums.yml
```diff
@@ -70,6 +70,10 @@
 #   pry>ENUMS::CAMPAIGN_STATUSES[2]          # active
 #
 
+BLOCK_LIST:
+  - https://cstream1.com/
+  - https://blockedurl.test/
+
```
hopsoft: Might be better to list the domains then parse the referrer and check the domain.

PR #859, opened 2019-12-02
Switch to heroku deflater

hopsoft:
# Enhancement

## Description

Switch to heroku-deflater for compression.

## Why should this be added

Smarter strategy for response compression.

### General

### CodeFund Core Team


Gemfile
```diff
@@ -29,6 +29,7 @@ gem "docraptor", "~> 1.3.0"
 gem "email_address", "~> 0.1.11"
 gem "full-name-splitter", "~> 0.1.2"
 gem "gibbon", "~> 3.3.0"
+gem "heroku-deflater", "~> 0.6.3", group: "production"
```
andrewmcodes: should we not do something like:

```suggestion
group :production do
  gem "heroku-deflater", "~> 0.6.3"
end
```

hopsoft: We'll switch to that format as soon as we have more than 1 production env only dep.

PR #868, opened 2019-12-04
Dependency improvements

andrewmcodes:
# Tweak CI config and webpacker improvements

## Description

Update circle config, add ability to analyze js bundle size, make some improvements based on that.

## Why should this be added

The js package I added will give us some more insights into the issues we are seeing with our js bundle. I also was able to decrease the size slightly, and also make an improvement to the circle config.

### General

### CodeFund Core Team

## Screenshots

Before:
image (3)

After:
image (2)


XhmikosR:
Should probably start importing only the needed modules :) Look into replacing moment with something else like date-fns, reduce lodash usage etc :)

.circleci/config.yml
```diff
@@ -9,7 +9,7 @@ jobs:
           RAILS_ENV: test
           RACK_ENV: test
           DEFAULT_HOST: codefund.io
-          PARALLEL_WORKERS: 1
+          PARALLEL_WORKERS: "1"
```
hopsoft: Odd that this needed to be a string.

PR #870, opened 2019-12-06
Remove identical conditions

XhmikosR:
Not sure if the second check was supposed to be something else, but right now it is identical.


andrewmcodes:
@all-contributors please add @XhmikosR for code

XhmikosR:
@andrewmcodes That's why we have tools to help us :-)

https://lgtm.com/projects/g/gitcoinco/code_fund_ads/?mode=list

IMO you guys should add it to the repo (it's owned by GH) and just disable commenting on PRs.

I'll try to fix the other issues later unless someone else beats me to it

PR #874, opened 2019-12-06
Update README.md

XhmikosR:
Fix the async attribute and remove the type since it's the default

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attr-async

BTW I'm not sure if there more instances like this, I will go through the codebase later.


XhmikosR:
We could also suggest `async defer` for browsers that don't support async, but my guess is you don't care about such browsers anyway :)

andrewmcodes:
@hopsoft we use this in our instructions for publishers. It looks like we need to change that.

hopsoft:
We originally had `async defer` but took defer out for a reason I can't recall now. I'm happy to recommend both again. Let's update both the readme and the instructions.

XhmikosR:
I'd say let's go with the modern way for now. I'm not sure where else the snippet is referenced; I'll do a scan later and update this PR or make another one

andrewmcodes:
Closing in favor of #1030

PR #875, opened 2019-12-06
Remove superfluous argument passed to anonymous function.

XhmikosR:


XhmikosR:
BTW these 2 files are identical, I'm not sure if it's possible to be moved to a common place to reduce duplication?

XhmikosR:
Why was this closed?

PR #902, opened 2019-12-11
Remove fail over expiration in attempt to avoid thundering herd

hopsoft:
# Fix attempt

## Description

We are seeing increase in pressure placed on the db at times which is seemingly
related to cache key expiration. This PR removes a failover key expiration which isn't needed
in an attempt to mitigate the behavior.

## Why should this be added

This should help alleviate observed intermittent latency on certain endpoints.

### General

### CodeFund Core Team


hopsoft:
We could also leave the fail over expirations but randomize them. Thoughts?

hopsoft:
Note that the cache keys already naturally expire as they include date and time block segments. See: https://github.com/gitcoinco/code_fund_ads/blob/master/app/lib/extensions/date_helpers.rb#L31-L38

PR #930, opened 2019-12-14
[WIP] Front App Integration

coderberry:
This is the beginning of setting up a plugin for the Front app. The goal is to use Stimulus Reflex to perform all actions and load necessary data.

I'm pretty excited about this one!


andrewmcodes:
I know you’re still working on this but this seems like the ideal use case for an engine 👀

coderberry:
I agree 100%. I even tried it but couldn’t dive deep enough.

andrewmcodes:
Is this something we are still wanting to do?

PR #938, opened 2019-12-16
Fix some issues with using current user vs current org

andrewmcodes:
# Bug Fix

## Description

I missed some controllers that needed to rely on current org not current user so that users can see details of campaigns etc. that are under the current org not user. Also fixes issue with the sidebar.

### General

### CodeFund Core Team


app/controllers/advertisement_previews_controller.rb
```diff
@@ -19,7 +19,7 @@ def show
     return render("/advertisement_previews/show") if request.format.html?
 
     # 2. Second it renders the JavaScript to preview the ad
-    @campaign ||= current_user.campaigns.build(
+    @campaign ||= Current&.organization&.campaigns&.build(
```
hopsoft: You'll also need to set the `user_id` on the built campaign I think.

app/views/shared/sidebar_content/_organizations.html.erb
```diff
@@ -1,32 +1,33 @@
...
+  "has-open": request.path.start_with?('/organization')
 ) do %>
   <a href="#" class="menu-link">
     <span class="menu-icon fad fa-users"></span>
     <span class="menu-text">Organizations</span>
   </a>
   <ul class="menu">
     <li class="menu-subhead">Organizations</li>
+    <%= active_link_to "Create Organization", new_organization_path, class: "menu-link", tabindex: -1, wrap_tag: :li, wrap_class: "menu-item" if authorized_user.can_admin_system? %>
+    <%# binding.pry %>
```
coderberry: We should remove this

PR #973, opened 2020-01-07
Add notification when a user is flagged by the CSL

andrewmcodes:
# Feature

## Description

Whenever a user is updated or created, run their name against the CSL to ensure they are not flagged.

Also fixes a small bug in the users form for users without an assigned org. 

Right now I am using a fork for `import_export` bc it does not work on the current version of rails and hasn't been touched for awhile but I am thinking of changing it further in the future and creating a whole new gem.

### General

### CodeFund Core Team


app/jobs/check_consolidated_screening_list_job.rb
```diff
@@ -0,0 +1,30 @@
+require "import_export"
+
+class CheckConsolidatedScreeningListJob < ApplicationJob
+  queue_as :low
+
+  def perform(user)
+    ScoutApm::Transaction.ignore! if rand > (ENV["SCOUT_SAMPLE_RATE"] || 1).to_f
```
hopsoft: The ignore logic for scout isn't really needed on low volume jobs.

PR #977, opened 2020-01-07
hide creatives edit button for active creatives

andrewmcodes:
# Bug fix

## Description

We want to keep the creatives edit button hidden if the creative is active and the authorized user is not an admin.

## Why should this be added

A user reached out to Eric about a 404 when updating their active creative and it was discovered that this button was visible when it should't be.

### General

### CodeFund Core Team


app/views/creatives/show.html.erb
```diff
@@ -3,7 +3,7 @@
   tabs: false,
   buttons: [
     layout_button(link: new_creative_path(clone: @creative.id), icon: "fad fa-copy", title: "Clone creative"),
-    layout_button(link: edit_creative_path(@creative), icon: "fad fa-pencil", title: "Edit creative"),
+    layout_button(link: edit_creative_path(@creative), icon: "fad fa-pencil", title: "Edit creative", admin: @creative.active?),
```
hopsoft: The `admin` argument is somewhat confusing. I wonder if we could think of better semantics for it.

andrewmcodes: I was thinking about that when I wrote this. I'll put it on my list.

PR #992, opened 2020-01-09
Remove the default scope on creatives

hopsoft:
# Fix

## Description

Remove the default scope on creatives.

## Why should this be added

The default scope is suspect in a performance drop and observed latency spikes.

### General

### CodeFund Core Team


app/models/concerns/impressionable.rb
```diff
@@ -137,7 +137,7 @@ def fetch_from_cache(key, force: false, &block)
       PgLock.new(name: key, attempts: 1).lock do
         value = block.call
         Rails.cache.write key, value, expires_in: 1.hour
-        Rails.cache.write fresh_key, true, expires_in: 10.minutes
+        Rails.cache.write fresh_key, true, expires_in: 7.minutes
```
andrewmcodes: Curious as to what impact this would have..

hopsoft: This will further reduce daily overspend.

PR #1016, opened 2020-01-16
paginator

### LINKED ISSUE #981: Simplify Pagination
# Feature Request

Currently, working with pagination is clunky. The window doesn't scroll to the top after you select a page. Also, the URL doesn't change so if I refresh a page and then go back, it won't go to the page I was on.

There are other issues related to this, but I'll create another issue for them.

I'd like to only use Stimulus Reflex when it makes sense to do so. I'm not sure this was the instance to use it. It might be better to revert pagination to a more conventional way.

## Describe alternatives you've considered

Bitch and moan

## Additional context

Discussed with @andrewmcodes

andrewmcodes:
# Enhancement

## Description

Update the pagination on data lists. Changing the page will now scroll you to the top, and you can go to another page, hit the back button, and be on the same page that you were in the list. 

Note the reason we reset the page position is we were concerned that it would effect your position on other lists but this is not the case.

Fixes #981

## Why should this be added

Fixes some UX concerns.

### General

### CodeFund Core Team


app/javascript/controllers/pagination_controller.js
```diff
@@ -0,0 +1,7 @@
+import ApplicationController from './application_controller'
+
+export default class extends ApplicationController {
+  beforeReflex() {
+    window.scrollTo(0, 0)
+  }
+}
```
hopsoft: Should this be in an `afterReflex` instead? Or does it work better in `before`?

andrewmcodes: It works better in before for whatever reason.

PR #1017, opened 2020-01-16
Optimize campaign query for ad render

hopsoft:
# Enhancement

## Description

Simplify the campaign query on ad render. Currently the campaign query includes a safeguard to ensure that the active campaign also has active creatives. We should be able to expect that this condition is always true which allows us to optimize ad rendering performance.

Note, we may need additional validation logic on creatives and campaigns to enforce the above expectation.

## Why should this be added

This should speed up ad rendering a little.

### General

### CodeFund Core Team


andrewmcodes:
Notes:
- Only active creatives are available for select on the campaigns form
- There are no active campaigns with non-active creatives

app/models/campaign.rb
```diff
@@ -425,6 +426,11 @@ def validate_creatives
     end
   end
 
+  def validate_active_creatives
+    return unless creatives
+    errors.add(:creatives, "cannot be inactive") unless creatives.map(&:active?).all?
```
hopsoft: I wonder if `creatives.active.exists?` would be better here.

PR #1021, opened 2020-01-16
Campaign bundles

hopsoft:
# Enhancement

## Description

Add the ability to easily build a campaign bundle.

## Why should this be added

Will make building and selling campaigns simpler and faster.

### General

### CodeFund Core Team


app/views/audiences/_table.html.erb
```diff
@@ -0,0 +1,26 @@
...
+              <span class="badge badge-subtle badge-primary"><%= keyword %></span>
+            <% end %>
+          <% end %>
+        <% end %>
+      <% end %>
+    <% else %>
+      <%= render "/shared/no_data_alert" %>
+    <% end %>
+  <% end %>
+<% end %>
```
hopsoft: Will be excited when this type of custom table builder can be moved to `ActionView::Component`

app/controllers/campaign_bundles_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+    @campaign_bundle.organization = Current.organization
+    @campaign_bundle.user = current_user
+    @campaign_bundle.campaigns.each do |campaign|
+      campaign.assign_attributes status: ENUMS::CAMPAIGN_STATUSES::ACCEPTED
+    end
+
+    respond_to do |format|
+      if @campaign_bundle.save
+        format.html { redirect_to campaign_bundles_path, notice: "Campaign bundle was successfully created." }
+        format.json { render :show, status: :created, location: @campaign_bundle }
```
andrewmcodes: Did we decide whether we needed this?

app/javascript/controllers/application_controller.js
```diff
@@ -1,18 +1,34 @@
...
-    document.getElementById('reflex-status').classList.add('d-none')
-    setFocus()
+  afterReflex (element) {
+    this.focus(element)
+  }
+
+  shouldFocus (element) {
+    if (document.activeElement === element) return false
+    if (element.tagName === 'TEXTAREA') return true
+    if (element.tagName !== 'INPUT') return false
```
andrewmcodes: we may want to add the date field input here. Until we add one of our date pickers, trying to type in a date on safari causes the focus issue.

app/javascript/controllers/select_date_controller.js
```diff
@@ -3,11 +3,22 @@
...
+    autoUpdateInput: false
+  })
+}
+
+document.addEventListener('cable-ready:after-morph', () => {
+  document
+    .querySelectorAll('[data-controller="select-date"]')
+    .forEach(el => init(el))
+})
+
```
andrewmcodes: are we leveraging any of the changes in this file?

app/javascript/controllers/text_field_controller.js
```diff
@@ -2,6 +2,7 @@ import { Controller } from 'stimulus'
 
 export default class extends Controller {
   connect () {
-    this.element.style.minHeight = `${window.innerHeight}px`
+    console.log('dispatching input', this.element)
+    this.element.dispatchEvent(new Event('input'))
```
andrewmcodes: Why do we need this?

app/models/campaign.rb
```diff
@@ -80,7 +116,19 @@ class Campaign < ApplicationRecord
   scope :paid_fallback, -> { where paid_fallback: true }
   scope :premium, -> { where(fallback: false).where(paid_fallback: false) }
   scope :job_posting, -> { where job_posting: true }
-  scope :available_on, ->(date) { where(arel_table[:start_date].lteq(date.to_date)).where(arel_table[:end_date].gteq(date.to_date)) }
+
+  scope :starts_on_or_before, ->(date) { where arel_table[:start_date].lteq(date.to_date) }
+  scope :starts_on_or_after, ->(date) { where arel_table[:start_date].gteq(date.to_date) }
```
andrewmcodes: I don't see this being used

app/models/campaign.rb
```diff
@@ -80,7 +116,19 @@ class Campaign < ApplicationRecord
...
+  scope :starts_after, ->(date) { where arel_table[:start_date].gt(date.to_date) }
+  scope :ends_on_or_before, ->(date) { where arel_table[:end_date].lteq(date.to_date) }
+  scope :ends_on_or_after, ->(date) { where arel_table[:end_date].gteq(date.to_date) }
+  scope :ends_after, ->(date) { where arel_table[:end_date].gt(date.to_date) }
```
andrewmcodes: I don't see this being used

app/models/concerns/campaigns/budgetable.rb
```diff
@@ -137,10 +138,17 @@ def hourly_budget_available?
 
     private
 
+    def init_total_budget
+      return unless daily_budget > 0
+      return unless total_budget == 0
+      self.total_budget = end_date ? total_operative_days * daily_budget : daily_budget * 365
```
andrewmcodes: I think we need to change this logic now that we will always have end dates.

app/views/layouts/demo.html.erb
```diff
@@ -12,6 +12,7 @@
     <%= content_for :demo %>
 
     <!-- CDN Stylesheets -->
+    <link rel="stylesheet" href="https://pro.fontawesome.com/releases/v5.11.2/css/all.css" integrity="sha384-zrnmn8R8KkWl12rAZFt4yKjxplaDaT7/EUkKm7AovijfrQItFWR7O/JJn4DAa/gx" crossorigin="anonymous" />
```
andrewmcodes: We should be able to remove this.

app/views/shared/_errors.html.erb
```diff
@@ -0,0 +1,13 @@
...
+      <% record.errors.full_messages.each do |message| %>
+        <li><%= message %></li>
+      <% end %>
+    </ul>
+  </div>
+<% else %>
+  <div id="error-explanation-<%= record.id || record.try(:temporary_id) %>" class="border border-success rounded p-3 mb-3 bg-white text-success">
+    Great job! Inputs are valid.
+  </div>
+<% end %>
```
andrewmcodes: Not sure how I feel about this - This is only for forms that aren't wired into simple form and we only have one form like that now..

andrewmcodes: I feel like this shouldn't be in shared but I'm not totally bothered by it

andrewmcodes:
sorting through these conflicts isn't going to be fun.

app/controllers/campaign_bundles_controller.rb
hopsoft: I think it's ok to keep for now.

app/javascript/controllers/application_controller.js
hopsoft: I think we can punt on that for now as a date picker will soon follow.

app/javascript/controllers/select_date_controller.js
hopsoft: No but it's a good refactor. I say we leave it.

app/javascript/controllers/text_field_controller.js
hopsoft: Good catch. We don't use this controller.

app/models/campaign.rb
hopsoft: I vote we keep these scopes even though they aren't being used in the codebase, I'll very likely be using them in a CLI.

app/models/campaign.rb
hopsoft: I vote we keep these scopes even though they aren't being used in the codebase, I'll very likely be using them in a CLI.

app/views/shared/_errors.html.erb
hopsoft: I vote keep and we can fork logic if we need it to support simple_form too at some point.

app/models/audience.rb
```diff
@@ -102,6 +103,35 @@ def ecpm_for_country_code(country_code)
...
+
+  def average_daily_impressions_counts(region: nil)
+    list = dailies(3.months.ago.beginning_of_month, 1.month.ago.end_of_month, region: region).to_a
+    (1..31).each_with_object({}) do |day, memo|
+      rows = list.select { |daily| daily.displayed_at_date.day == day }
+      average_impressions_count = (rows.map(&:impressions_count).sum / rows.size.to_f).round
+      memo[day] = average_impressions_count
+    end
+  end
+
```
hopsoft: Pretty sure that these methods can be removed as I moved these concerns to Country and Region.

PR #1024, opened 2020-01-16
add ability to automatically trigger organization report delivery

### LINKED ISSUE #991: Add button to trigger organization report delivery
# Feature Request

Currently for me to send an organization report, I have to go to organization > reports and build it by hand. The information for generating these reports is always the same:

* Select active campaigns
* Set the start/end dates to the earliest start date and last end date of the selected campaigns
* Set the title to '___Team Name___ Summary Report'
* Set the recipients to members of the team and myself

This is tedious and can be automated.

I would like to see a button on the teams index page and the organization shared layout headers called "Generate Reports" that triggers the above into a background job (see generate_organization_report_job.rb).

andrewmcodes:
# Feature

## Description

Add ability to automatically trigger organization report delivery.

Process:
- Select active campaigns
- Set the start/end dates to the earliest start date and last end date of the selected campaigns
- Set the title to '_Team Name_ Summary Report'
- Set the recipients to members of the team and Eric

Fixes #991

## Why should this be added

This will save Eric a lot of time.

### General

### CodeFund Core Team


app/jobs/create_organization_report_job.rb
```diff
@@ -0,0 +1,33 @@
+class CreateOrganizationReportJob < ApplicationJob
+  queue_as :default
+
+  def perform(id:)
+    organization = Organization.find(id)
+    return unless organization
```
hopsoft: `find` will raise an error if we don't find the record. You may want to change to `find_by id: id` so this job doesn't fail and reenque.

app/jobs/create_organization_report_job.rb
```diff
@@ -0,0 +1,33 @@
...
+    campaigns = organization.campaigns&.active
```
hopsoft: Safe navigation not needed since we exit when `organization` doesn't exist.

app/jobs/create_organization_report_job.rb
```diff
@@ -0,0 +1,33 @@
...
+    return unless campaigns
```
hopsoft: This will be an instance of an ActiveRecord Relation. Probably better to check `campaigns.exists?` for early exit.

Issue #1034, opened 2020-01-19
Behaviour of JS callback with fallback disabled


JasonBarnabe:
https://github.com/gitcoinco/code_fund_ads#embed-callbacks

What is `evt.detail` set to if there are no paid ads available and I have disabled fallback ads?

(Please document this.)

coderberry:
@hopsoft or @andrewmcodes this should be a quick addition (and needed)

andrewmcodes:
@JasonBarnabe this is the last thing in the embed-callbacks section.

> And in the event that we do not have an available advertiser, you will see:
> 
> `{ "status": "no-advertiser" }`

I recreated your situation and used this markup:

```html
<!DOCTYPE html>
<html>
  <body>
    <!-- BEGIN COPY CODEFUND EMBED -->
    <div id="codefund"></div>
    <script
      src="https://app.codefund.io/properties/###/funder.js"
      async
    ></script>
    <!-- END COPY CODEFUND EMBED -->

    <script>
      window.addEventListener("codefund", function (evt) {
        if (evt.detail.status !== 'ok') {
          console.log(evt.detail); // {status: "no-advertiser"}
          console.log(evt.detail.status); // no-advertiser
        }
      });
    </script>
  </body>
</html>
```

So if if there are no paid ads available and you have disabled fallback ads, `evt.detail` will be `{status: "no-advertiser"}`. 

If the documentation in the README is unclear, I would love some suggestions for how to improve it!

JasonBarnabe:
Thanks for the answer.

The documentation is unclear because "we do not have an available advertiser" is ambiguous. Is a fallback ad an "available advertiser"? If so, does turning off fallback ads make them "unavailable"? Are there scenarios where you'd get "no-advertiser" even if fallback ads are on?

I would suggest rewording to something like:

> And in the event we have no ad to show (no paid ad, and you've turned off fallback ads), you will see:

andrewmcodes:
@JasonBarnabe thanks, I will make an update to the wording. 

No advertiser means there is no available paid or fallback advertiser so I will make sure that it becomes more clear.

PR #1041, opened 2020-01-22
Update embed callbacks documentation

andrewmcodes:
# Enhancement

## Description

Update the documentation to be a bit more clear about what happens when we do not have an available advertiser based on feedback in #1034 when using embed callbacks.

It also occurs to me that some of the return values of `evt.detail` are a little confusing. 

```json
// Paid
{ "status": "ok", "house": false }
// Fallback
{ "status": "ok", "house": true } 
// Error
{ "status": "error", "message": "error message" }
// No available ads (fallback or paid)
{ "status": "no-advertiser" }
```

Personally I think this should look more like: 
```json
// Paid
{ "status": "ok", "fallback": false }
// Fallback
{ "status": "ok", "fallback": true } 
// Error
{ "status": "error", "message": "error message" }
// No available ads (fallback or paid)
{ "status": "no-ad" } // I don't love this being the status but ¯\_(ツ)_/¯ 
```

### General

### CodeFund Core Team


hopsoft:
Agree that the design of this feature isn't great. I'd prefer to see something like:

```js
{ status: 'ok', type: 'paid' } // HTTP 200
{ status: 'ok', type: 'unpaid' } // HTTP 200
{ status: 'not found', type: 'none' } // HTTP 404
{ status: 'error', message: 'error message' } // HTTP 500
```

_Note that the HTTP status codes already return the proper status codes._

andrewmcodes:
@hopsoft we can add that to the backlog and I will get to it.

PR #1069, opened 2020-02-05
fix 'this month' selection for datepicker

### LINKED ISSUE #1058: Publisher date filter presets do not update as days pass
Today, if I choose "This Month", I will get a filter for Jan 1 - 26. Tomorrow, if I load the dashboard again without touching the date filters...

Expected behaviour: filter is set to Jan 1 - 27.

Current behaviour: filter is set to Jan 1 - 26; I have to pick "This Month" again to get it to update.

andrewmcodes:
# Bug fix

## Description

Fix logic error when selecting 'This Month' on date range selector.

Fixes #1058

## Why should this be added

Currently, when a user hits 'This Month' on the date range selector, it selects from the beginning of the month to today. This should actually be selecting the beginning of this month to the end of this month.

### General

### CodeFund Core Team


app/javascript/controllers/dashboard_date_range_controller.js
```diff
@@ -53,7 +53,7 @@ export default class extends Controller {
               .add(-1, 'month')
               .endOf('month')
           ],
-          'This Month': [moment().startOf('month'), moment()],
+          'This Month': [moment().startOf('month'), moment().endOf('month')],
```
hopsoft: Not sure this will address the end date stored in session.

andrewmcodes: It should - happy to elaborate.

PR #1070, opened 2020-02-05
Add ability for advertisers to pause or unpause campaigns

### LINKED ISSUE #1060: Allow advertisers to toggle the status of their active campaigns
# User Story

As an advertiser I want be able to pause active campaigns so that I do not spend money while waiting for an administrator to pause them.

## Description

Advertisers need the ability to pause and unpause their campaigns.

## Acceptance Critera

1. Advertisers can pause active campaigns
2. Advertisers can start paused campaigns that were initially started by an admin

coderberry:
# Feature

## Description

As an advertiser, I can pause and continue active or paused campaigns.

Fixes #1060 

## Why should this be added

It's critical for advertisers to be able to pause active campaigns in order to make changes or delay the spend.

### General

### CodeFund Core Team

## Screenshots

CodeFund___Ethical_Advertising_-_Firefox_Developer_Edition


app/helpers/frontend/button_helper.rb
```diff
@@ -24,7 +24,8 @@ def generate_organization_report_button(id)
       button_tag("Generate Reports", data: {reflex: "click->OrganizationReportsReflex#generate", id: id, controller: "organization-reports"}, class: "ml-1 btn btn-primary", style: "height:2.5rem;")
     end
 
-    def layout_button(link: "#", icon: nil, add_class: "", target: "", title: nil, admin: false, method: :get, confirmation_message: nil, permissions: true)
+    def layout_button(link: "#", icon: nil, add_class: "", target: "", title: nil, admin: false, method: :get, confirmation_message: nil, permissions: true, display: true)
```
andrewmcodes: This is getting a bit unwieldy. Permissions and display function the same. I like display over permissions. Could we remove permissions and replace it with display in the views? It isn’t used much.

app/controllers/campaigns_controller.rb
```diff
@@ -140,7 +140,10 @@ def advertiser_campaign_params
       :name,
       :url,
       creative_ids: [],
-    )
+    ).tap do |whitelisted|
+      new_status = params[:campaign][:status]
+      whitelisted[:status] = new_status if ["paused", "active"].include?(new_status) && ["paused", "active"].include?(@campaign.status)
+    end
```
andrewmcodes: I’m not really a fan of this. Could we abstract this out into an `authorize_` method (see properties controller)? That way we could flash a descriptive error message.

hopsoft: 100% agree on moving this logic to an authorizer.

app/helpers/frontend/button_helper.rb
hopsoft: I'm not really up to date on what `layout_button` is doing but agree that `permissions: true` is ambiguous. `display:true` is certainly more clear about intention and behavior.

hopsoft:
I agree with using campaign status for this behavior, but think we need to clearly define the semantics around each. What it means, what it does, business rules, etc...

- **pending** 

  - Not approved
  - May never become **active**
  - Not fully configured *i.e. missing creatives etc...*
  - Not included when determining sold inventory

- **active**

  - Approved
  - Fully configured
  - Available to show ads based on start and end dates
  - Included when determining sold inventory

- **paused**

  - Approved
  - Fully configured
  - Will not show ads
  - Included when determining sold inventory
  - Only **active** campaigns can be put into this state

- **archived**

  - No longer in use
  - Not included when determining sold inventory

---

This seems sound; however, we still have a use case that isn't addressed by these status definitions.

Consider the new campaign bundle builder. Campaigns created this way don't yet have their creatives defined. Also, we haven't yet received payment but need to include these campaigns when determining sold inventory. I initially thought we'd use the **paused** status for this concern but the existing rules prevent this without changing semantics.

Do we want to change the semantics around **paused** or do we want to add a new status? A new status would look something like this.

- **sold**

  - Approved
  - Will become **active**
  - Not fully configured *i.e. missing creatives etc...*
  - Included when determining sold inventory

microftech65:
Add ability for advertisers to pause or unpause campaigns#1070

PR #1071, opened 2020-02-06
update the csl gem

andrewmcodes:
# Feature

## Description

The Trade.gov API went through a massive change. This change will fix the issues that those changes created.

### General

### CodeFund Core Team


app/jobs/check_consolidated_screening_list_job.rb
```diff
@@ -1,4 +1,4 @@
-require "import_export"
+require "consolidated_screening_list"
```
hopsoft: I the explicit require actually needed?

app/jobs/check_consolidated_screening_list_job.rb
```diff
@@ -7,8 +7,8 @@ def perform(user)
     ScoutApm::Transaction.ignore! if rand > (ENV["SCOUT_SAMPLE_RATE"] || 1).to_f
     @user = user
     return unless @user
-
-    return if ImportExport::Client.new.search(name: @user.name).empty?
+    client = ConsolidatedScreeningList::Client.new api_key: ENV["TRADE_API_KEY"]
```
hopsoft: This environment variable may be a little too generic. Thoughts?

andrewmcodes: This is what they recommend in their docs. Its the api key for Trade.gov so I think we should keep it.

PR #1078, opened 2020-02-07
Refactor inventory summary/details

hopsoft:


app/models/campaign.rb
```diff
@@ -277,12 +277,13 @@ def to_stashable_attributes
...
   def inventory_summary
-    return nil if persisted? || campaign_bundle.nil?
     @inventory_summary ||= InventorySummary.new(self)
   end
 
   def targeting_variants
-    regions.to_a.product audiences.to_a
+    r = region_ids.present? ? regions : regions(include_fuzzy_matches: true)
+    a = audience_ids.present? ? audiences : audiences(include_fuzzy_matches: true)
+    r.to_a.product a.to_a
```
andrewmcodes: 🤢

hopsoft: Required to have this work with both ways of building campaigns.

PR #1094, opened 2020-02-16
Frontend cleanup

### LINKED ISSUE #932: Changes/diff modal and created/updated at
The modal that shows when you click "x made changes" on the property dashboard...

- Always shows "created at" with "No Changes".
- Always shows "updated at" as changed.

Neither of these should show.

andrewmcodes:
# Enhancement

## Description

- Add in pagination on view tables it wasn't being used on
- Several improvements to creatives index
- improvements to reflex pagination
- lot of cleanup to the helpers

Fixes #932

### General

### CodeFund Core Team


app/controllers/campaign_dashboards_controller.rb
```diff
@@ -1,3 +1,5 @@
+# TODO: not being used
+
```
hopsoft: Should we update with a deprecation?

app/controllers/images_controller.rb
```diff
@@ -6,7 +6,7 @@ class ImagesController < ApplicationController
   before_action :authorize_edit!, only: [:edit, :update]
 
   def index
-    @images = @imageable.images
+    @images = @imageable.images.includes([:blob])
```
hopsoft: I think you can omit the array in the includes.

app/controllers/campaign_dashboards_controller.rb
andrewmcodes: Only reason I didn't is bc I wanted to talk to Eric about it

PR #1099, opened 2020-02-19
fix issues with blank fields on campaign form

andrewmcodes:
# Bug fix

## Description

Fix issue with blank fields on campaign form after save.

## Why should this be added

The creatives issue was just caching but the issue with the user id not showing up seemed to be because of the sync to campaign bundle method. I don't see a reason why we need to sync this field, especially since the person who creates the bundle will be eric for the near future.

### General

### CodeFund Core Team


app/views/campaigns/_form.html.erb
```diff
@@ -183,7 +183,7 @@
         <small class="mt-1 d-block text-primary opacity-70">Multivariate testing is used to determine the most effective creative</small>
       <% end %>
     <% end %>
-    <% cache Current.organization.creatives.active do %>
+    <% cache "#{Current.organization.creatives.active}-#{campaign.creative_ids}" do %>
```
hopsoft: I would change this to the following

```erb
<% cache [Current.organization.creatives.active, campaign.creative_ids.join] do %>
```

PR #1109, opened 2020-02-24
Add back pressure check on ad render

hopsoft:
# Fix

## Description

A safety measure to prevent overspend when the system is experiencing back pressure.

## Why should this be added

The platform is essentially a distributed system with eventual data consistency given all the async data processing that happens. This leaves us at risk of overspend when we experience back pressure i.e. too much latency when recording impressions. This PR eliminates the risk of significant overspend whenever this happens.

### General

### CodeFund Core Team


config/initializers/rack_attack.rb
```diff
@@ -21,12 +21,12 @@
 
 if Rails.env.production?
   # Throttle all IPs to 20 requests/minute
-  Rack::Attack.throttle("requests by ip", limit: 20, period: 1.minute.to_i) do |request|
+  Rack::Attack.throttle("requests by ip", limit: ENV.fetch("MAX_REQUESTS_PER_IP_PER_MIN", 20).to_i, period: 1.minute.to_i) do |request|
```
andrewmcodes: Can you add these to the example env?

PR #1119, opened 2020-02-27
Change to a single impressions query for daily summary rollup

hopsoft:
# Enhancement

## Description

Currently the daily summary job executes 7 queries against the impressions table. This reduces that to a single query. Note that it will put more pressure on memory consumption for this job, but should relieve some of the pressure on the db.

### General

### CodeFund Core Team


app/models/impression.rb
```diff
@@ -34,11 +34,12 @@ class Impression < ApplicationRecord
...
   # relationships .............................................................
-  belongs_to :advertiser, class_name: "User", foreign_key: "advertiser_id"
-  belongs_to :publisher, class_name: "User", foreign_key: "publisher_id"
-  belongs_to :campaign
+  belongs_to :organization, optional: true
+  belongs_to :advertiser, class_name: "User", foreign_key: "advertiser_id", optional: true
+  belongs_to :publisher, class_name: "User", foreign_key: "publisher_id", optional: true
+  belongs_to :campaign, optional: true
   belongs_to :creative, optional: true
-  belongs_to :property
```
andrewmcodes: Why would this be optional?

config/initializers/money.rb
```diff
@@ -1,2 +1,3 @@
 Money.default_currency = Money::Currency.new("USD")
 Money.locale_backend = :currency
+Money.rounding_mode = BigDecimal::ROUND_HALF_EVEN
```
andrewmcodes: 🙌

config/sidekiq_data.yml
```diff
@@ -1,5 +1,8 @@
 :concurrency: <%= ENV.fetch("REDIS_QUEUE_MAX_THREADS", 5).to_i - 2 %>
 :queues:
+  - create_debits_for_campaign_and_date
+  - create_debit_for_campaign_and_date
+  - recalculate_organizataion_balances
   - ensure_daily_summaries
   - daily_summaries
   - daily_summary
```
andrewmcodes: Does anything need to be removed from this list?

app/models/impression.rb
hopsoft: I made these all optional to support more flexibility with the seeds and to prevent impressions from blocking other operations for the data they reference.

config/sidekiq_data.yml
hopsoft: Nope all are required and very explicit to better control the order of operations and priority.

PR #1127, opened 2020-03-02
Fix paper cuts (annoying stuff)

coderberry:
# Type of PR (feature, enhancement, bug fix, etc.)

## Description

* Add Metabase dashboard to Organization show page
* Default org report recipients to all listed in members tab
* Fix default sort direction of org transactions
* Update datepicker to go into past, not future
* Use the admin dashboard table for campaign list views

Fixes #(issue)

## Why should this be added

Because it solves a lot of annoyances in the UX

### General

### CodeFund Core Team


app/controllers/campaigns_controller.rb
```diff
@@ -12,8 +12,8 @@ class CampaignsController < ApplicationController
 
   def index
     campaigns = scope_list(Campaign)
-      .order(order_by)
-      .includes(:user, :creative, :organization)
+      .order(end_date: :desc)
+      .includes(:organization)
```
hopsoft: Pretty sure the explicit `includes` for `user` and `creative` was from an effort to squash n+1 queries. Are you certain that we haven't reintroduced some with this change?

coderberry: I don't believe they are being used on the new campaign list

PR #1148, opened 2020-03-10
Update our sorting/filtering pattern

andrewmcodes:
# Enhancement

## Description

- Establish clearer pattern for sorting/filtering
- Update SR
- Update pagy and fix associated TODO's

## Why should this be added

Creates a much clearer pattern that is also faster.

### General

### CodeFund Core Team


app/controllers/application_controller.rb
```diff
@@ -17,10 +17,17 @@ class ApplicationController < ActionController::Base
...
   impersonates :user
 
   protected
 
+  def set_last_controller_name
+    unless controller_name.include? "sparkline"
+      session[:last_controller_name] = controller_name
+    end
+  end
+
```
hopsoft: Are we still using `last_controller_name` anywhere?

app/controllers/concerns/scopable.rb
```diff
@@ -2,26 +2,16 @@ module Scopable
...
 
   included do
-    helper_method :scope_list
-    before_action :reset_current_scope
+    before_action :set_scoped_by
   end
 
   def scope_list(model)
-    model.public_send(current_scope)
+    model.public_send @scoped_by
```
hopsoft: We should probably add a TODO to tighten up security on this.

app/javascript/controllers/index.js
```diff
@@ -4,8 +4,9 @@
 import { Application } from 'stimulus'
 import { definitionsFromContext } from 'stimulus/webpack-helpers'
 import StimulusReflex from 'stimulus_reflex'
+import consumer from '../channels/consumer'
 
 const application = Application.start()
 const context = require.context('controllers', true, /_controller\.js$/)
 application.load(definitionsFromContext(context))
-StimulusReflex.initialize(application)
+StimulusReflex.initialize(application, { consumer })
```
hopsoft: Nice.

app/controllers/application_controller.rb
andrewmcodes: good catch

app/controllers/concerns/scopable.rb
```diff
@@ -2,16 +2,22 @@ module Scopable
...
   protected
 
   def set_scoped_by
     @scoped_by ||= "all"
   end
+
+  # Abstract method that should be overridden in including controllers
+  def scopable_values
+    raise NotImplementedError, "controller should implement `scopable_values`"
+  end
```
hopsoft: Shouldn't this method be named `set_scopable_values`?

PR #1157, opened 2020-03-12
Update start/end date syncing strategy between bundles and campaigns

andrewmcodes:
# Bug fix

## Description

Fix issue where saved campaigns were having their date range change reverted for bundled campaigns.

### General

### CodeFund Core Team

## Screenshots

CodeFund _ Ethical Advertising 2020-03-12 16-32-18


app/models/campaign.rb
```diff
@@ -619,8 +619,8 @@ def sync_to_campaign_bundle
     return unless campaign_bundle
     self.organization_id = campaign_bundle.organization_id
     self.region_ids = campaign_bundle.region_ids
-    self.start_date = campaign_bundle.start_date
-    self.end_date = campaign_bundle.end_date
+    campaign_bundle.update(start_date: start_date) if start_date < campaign_bundle.start_date
+    campaign_bundle.update(end_date: end_date) if end_date > campaign_bundle.end_date
```
hopsoft: I might try to condense this to a single db query. We may also want to set a flag here and only update the bundle on save rather than validate.

app/models/campaign.rb
```diff
@@ -619,9 +620,17 @@ def sync_to_campaign_bundle
...
+
+  def update_campaign_bundle_dates
+    return unless campaign_bundle
+    params = {
+      start_date: start_date < campaign_bundle.start_date ? start_date : nil,
+      end_date: end_date > campaign_bundle.end_date ? end_date : nil
+    }.compact
+    return if params.empty?
+    campaign_bundle.update(params)
+  end
```
hopsoft: Use of `Hash#compact` 🥰

test/integration/advertisements_test.rb
```diff
@@ -5,7 +5,7 @@ class AdvertisementsTest < ActionDispatch::IntegrationTest
   setup do
     Rails.cache.clear
     start_date = Date.parse("2019-01-01")
-    @premium_campaign = campaigns(:premium_bundled)
+    @premium_campaign = amend(campaigns: :premium_bundled, start_date: start_date, end_date: start_date.advance(months: 3))
```
andrewmcodes: Scared to merge this bc im not sure what you were testing here. I assume it may just be that the start and end date were correctly syncing from the bundle to campaign when the campaign is saved but I want to make sure.

Since the bundled campaign is no longer syncing its start and end dates with the campaign bundle, the campaign we create here, the properties can't find it bc it's in the future from where they think time is currently.

test/models/campaign_test.rb
```diff
@@ -258,4 +258,13 @@ class CampaignTest < ActiveSupport::TestCase
...
   end
+
+  test "bundled start and end date changes will update bundle" do
+    bundled_campaign = campaigns(:premium_bundled)
+    assert_equal bundled_campaign.campaign_bundle.start_date, bundled_campaign.start_date
+    assert_equal bundled_campaign.campaign_bundle.end_date, bundled_campaign.end_date
+    bundled_campaign.update(start_date: Date.parse("2020-03-01"), end_date: Date.parse("2020-05-01"))
+    assert_equal bundled_campaign.start_date, bundled_campaign.campaign_bundle.start_date
+    assert_equal bundled_campaign.end_date, bundled_campaign.campaign_bundle.end_date
+  end
```
andrewmcodes: bleh

test/integration/advertisements_test.rb
hopsoft: These tests were written before the system tests and are just verifying that the controller emits the expected output. Feel free to tweak test setup as needed for this new behavior on bundled campaigns.

andrewmcodes: ok then this should be good to merge.

PR #1190, opened 2020-04-01
Change how we are pulling the maxmind db

andrewmcodes:
# Enhancement

## Description

Update the way we are pulling in the maxmind db on intialization and set fallback if the file doesnt download.

## Why should this be added

Maxmind changed their license.

### General

### CodeFund Core Team

- [ ] I have requested a review from another developer(s)


config/initializers/maxmind.rb
```diff
@@ -1 +1 @@
-DownloadAndExtractMaxmindFileJob.new.extract
+DownloadAndExtractMaxmindFileJob.new.perform
```
hopsoft, line 1: Thinking out loud here, but we'd see faster boot times if we simply let this run as a background job. Trying to think through the implications of that. 🤔

hopsoft: Maybe we only run it synchronously in production. Thoughts?

config/initializers/maxmind.rb
```diff
@@ -1 +1,5 @@
-DownloadAndExtractMaxmindFileJob.new.extract
+if Rails.env.production? || ENV["CI"] == "true"
+  DownloadAndExtractMaxmindFileJob.new.perform
```
hopsoft, line 2: `perform_now` might be a more idiomatic Railsism for this but not a big deal.

PR #1209, opened 2020-04-08
Update the block advertiser feature to use organization_id rather than user_id

andrewmcodes:
# Feature

## Description

Adds the ability to block an organization from advertising on a property instead of a specific advertiser.

When this is merged, we need to run `rails properties:backfill_prohibited_org_ids` to update properties with the correct prohibited orgs based on their prohibited advertisers. Then #1210 needs to be merged and deployed.

Fixes #1198 & #198

### General

- [x] My code follows the style guidelines of this project & the tests are passing
- [x] I have performed a self-review of my own code
- [x] I have updated DB schema annotations or I have not made any DB schema changes

### CodeFund Core Team

- [x] I have added the appropriate milestone if applicable
- [x] I have requested a review from another developer(s)

## Screenshots

image

PR #1210, opened 2020-04-08
drop prohibited_advertiser_ids from properties

andrewmcodes:
# Enhancement

## Description

Cleanup from #1209. Drops `prohibited_advertiser_ids` from properties table and removes some remaining references to it.

### General

### CodeFund Core Team


app/models/property.rb
```diff
@@ -104,7 +104,7 @@ class Property < ApplicationRecord
   scope :for_campaign, ->(campaign) {
     relation = active.with_any_keywords(*campaign.keywords).without_any_keywords(*campaign.negative_keywords)
     relation = relation.where(prohibit_fallback_campaigns: false) if campaign.fallback?
-    relation = relation.without_all_prohibited_advertiser_ids(campaign.id)
+    relation = relation.without_all_prohibited_organization_ids(campaign.id)
```
hopsoft, line 107: The argument here looks wrong. FWIW, I think it was wrong before. It should be:

```ruby
relation = relation.without_all_prohibited_organization_ids(campaign.organization_id)
```

db/migrate/20200422185634_add_prohibited_organization_ids_to_properties.rb
```diff
@@ -0,0 +1,6 @@
+class AddProhibitedOrganizationIdsToProperties < ActiveRecord::Migration[6.0]
+  def change
+    add_column :properties, :prohibited_organization_ids, :bigint, default: [], null: false, array: true
+    add_index :properties, :prohibited_organization_ids, using: :gin
+  end
+end
```
hopsoft, line 6: What is our plan to migrate the existing data into this new column?

db/migrate/20200422190916_remove_prohibited_advertiser_ids_from_properties.rb
```diff
@@ -0,0 +1,10 @@
+class RemoveProhibitedAdvertiserIdsFromProperties < ActiveRecord::Migration[6.0]
+  def up
+    remove_column :properties, :prohibited_advertiser_ids
+  end
+
+  def down
+    add_column :properties, :prohibited_advertiser_ids, :bigint, default: [], null: false, array: true
+    add_index :properties, :prohibited_advertiser_ids, using: :gin
+  end
+end
```
hopsoft, line 10: This will need to be a phased deploy where we only drop this column after the data migration has been run successfully.

db/migrate/20200422185634_add_prohibited_organization_ids_to_properties.rb
hopsoft: Ignore as this is addressed by #1209

db/migrate/20200422190916_remove_prohibited_advertiser_ids_from_properties.rb
hopsoft: Ignore as this is addressed by #1209

PR #1229, opened 2020-04-21
Track fallback clicks on daily summaries table

coderberry:
# Enhancement

## Description

This PR will allow the storing of fallback clicks in the daily_summaries table. This will allow for much faster reporting.

### General

- [x]  I have performed a self-review of my own code

### CodeFund Core Team

## Screenshots

N/A


app/jobs/create_daily_summary_job.rb
```diff
@@ -1,4 +1,4 @@
-class CreateDailySummaryJob < ApplicationJob
+aclass CreateDailySummaryJob < ApplicationJob
```
coderberry, line 1: typo

Gemfile
```diff
@@ -52,6 +52,7 @@ gem "pg", ">= 0.18", "< 2.0"
 gem "premailer-rails", "1.10.3"
 gem "premailer", "1.11.1"
 gem "pretender", "~> 0.3.3"
+gem "progressbar", "~> 1.10.1"
```
andrewmcodes, line 55: This library is kinda big. Could we use [tty-progressbar](https://github.com/piotrmurach/tty-progressbar) Or do require false here since you have an explicit require in the task?

bin/copy_production_db_to_local
```diff
@@ -3,6 +3,6 @@ source .env
 export RAILS_ENV=development
 bundle exec rails db:migrate db:drop db:create
 ./bin/heroku_pg_dump_production_replica
-pg_restore --verbose --clean --no-acl --no-owner -h localhost -d code_fund_ads_development tmp/production-shallow.dump
+pg_restore -U $PG_USERNAME --verbose --clean --no-acl --no-owner -h localhost -d code_fund_ads_development tmp/production-shallow.dump
```
andrewmcodes, line 6: Is this in the env-example file?

Gemfile
coderberry: Good call. I'll use `require: false`

bin/copy_production_db_to_local
coderberry: I will add it there as well. Good call

coderberry: Yes it is

PR #1231, opened 2020-04-22
Fix creative validations

andrewmcodes:
# Bug fix

## Description

Correctly validate that images have been uploaded with a creative due to the way we assign images after the object is created/updated.

I'm not overjoyed by moving the image validations farther from the database but I think this is the best solution given the way we assign images.

Also adds a few enhancements to the creatives form.

## Why should this be added

#1227 introduced some issues that I missed.

### General

### CodeFund Core Team

## Screenshots

screencapture-localhost-3000-creatives-2020-04-22-13_17_39


app/controllers/creatives_controller.rb
```diff
@@ -29,13 +29,15 @@ def new
 
   def create
     @creative = current_user.creatives.build(creative_params)
+    images_valid = creative_image_params_validation.valid?
 
     respond_to do |format|
-      if @creative.save
+      if @creative.save && images_valid
```
andrewmcodes, line 35: Need to do it like this bc if we call `creative_image_params_validation.valid?` on this line, and there are errors that prevent the creative from saving, it won't call `.valid?` which means the errors will not get added to the object which would make `creative_image_params_validation.errors` empty. Similarly, if we were to do `creative_image_params_validation.valid? && @creative.save`, any errors on the creative itself would not be added to the error messages.

app/models/creative.rb
```diff
@@ -48,12 +48,12 @@ class Creative < ApplicationRecord
...
   validates :status, inclusion: {in: ENUMS::CREATIVE_STATUSES.values}
   validates :creative_type, inclusion: {in: ENUMS::CREATIVE_TYPES.values}
-  validate :validate_icon_image
-  validate :validate_small_image
-  validate :validate_wide_image
-  validate :validate_large_image
+  validates :icon_image, presence: {on: :update, message: "must be selected"}
+  validates :small_image, presence: {on: :update, message: "must be selected"}
+  validates :wide_image, presence: {on: :update, message: "must be selected"}
+  validates :large_image, presence: {on: :update, message: "must be selected"}
```
andrewmcodes, line 54: only checking on update here since the images won't be present on create.

PR #1234, opened 2020-04-23
don't load ads stylesheet when the theme is `unstyled`

### LINKED ISSUE #1228: Don't load CSS file when theme = unstyled
# Feature Request

I use unstyled CodeFund ads for performance reasons. However, it still loads the CSS file even when the theme is unstyled. 0% of the CSS in the file is used.

Screen Shot 2020-04-21 at 5 50 30 pm

Remove the stylesheet link from the HTML when the theme is `unstyled`

## Describe alternatives you've considered

For the moment, I've removed the stylesheet from the HTML manually with a regular expression before adding to the DOM:
```javascript
fetch("https://api.codefund.app/properties/xx/funder.html?$third-party")
  .then(r => r.text())
  .then(
    html =>
      (document.getElementById("codefund").innerHTML = html
        /* Remove stylesheet link to save network usage */
        .replace(/<link .*?rel="stylesheet".*?>/g, ""))
  )
  .catch(
    () =>
      (document.getElementById("codefund").innerHTML =
        '<div class="codefund-placeholder" style="text-align: left">Please consider disabling your ad blocker to support this site.<br><br>This rectangle is the only place on the page where ads are shown.</div>')
  );
```

## Additional context

I am using the `funder.html` way of doing things rather than `funder.js` (again for performance reasons).

andrewmcodes:
## Type of PR

- [ ] feature
- [x] enhancement
- [ ] bug fix
- [ ] other

## Description

There is no reason to include the stylesheet when we render the ad if the theme is `unstyled`. This change will help publishers who are using the `unstyled` theme to save time on each request by not downloading the CSS.

Fixes #1228

## Checklist


app/views/advertisements/show.js.erb
```diff
@@ -12,10 +12,11 @@
     new CodeFundAd(<%= advertisement_mustache_template_options.to_json.html_safe %>)
   }
 
+  const codefundThemeName = '<%= theme_name.to_s %>'
   const codefundStylesheetId = 'codefund-style'
   const codefundScriptId = 'codefund-script'
 
-  if (!document.getElementById(codefundStylesheetId)) {
+  if (!document.getElementById(codefundStylesheetId) && codefundThemeName != 'unstyled') {
```
hopsoft, line 19: I would use `codefundThemeName !== 'unstyled'` here.

Booligoosh:
Wow, thanks for the quick resolution!

Unfortunately, this doesn't work with the funder.html pages (eg. https://api.codefund.app/properties/608/funder.html) as far as I can tell.

I'm fairly sure that making this change would fix it:
https://github.com/gitcoinco/code_fund_ads/commit/2ba541e04042f653d066f1a3c66b9dfef23f904f

I've committed it on a forked branch from the GitHub web interface, so I didn't submit a PR yet since I haven't downloaded and set up the entire project on my machine yet so couldn't run tests and verify that it works.

However, do you think I'm on the right track, and do you think that would be something that you would be interested in accepting? If so, I can make a PR.

PR #1236, opened 2020-04-27
Include callbacks in demo (for demonstration)

### LINKED ISSUE #1196: CodeFund JS event sometimes fires twice
I use an event listener so that when CodeFund has no paid ad, I put in another ad network in its place. My code is [here](https://github.com/JasonBarnabe/greasyfork/blob/master/app/assets/javascripts/ads.js).

I noticed that sometimes the event fires twice. This is messing up my ad swapping (though I can fix that easily enough).

The page serving the ad is [here](https://greasyfork.org/en/scripts/375583-killermod-all-in-one-the-best-moomoo-io-mod-hack-spike-trap-hotkey-autobull-autoheal-and-more). When the problem happens, you can see this in the console:

```
CodeFund does not have an advertiser for you at this time.
codefund event: [object Object]
CodeFund container element not visible! Please verify an element exists (and is visible) that matches the CSS selector "#codefund_ad"
codefund event: [object Object]
```

(Followed by some errors caused by my assumption this would only run once.)

This problem does not occur on every page load. I'm testing with Firefox.

coderberry:
## Type of PR

- [ ] feature
- [ ] enhancement
- [ ] bug fix
- [x] other

## Description

- Add demonstration of using CodeFund callbacks with the Demo landing page.
- Add quick preview link for advertisers to view their ads 

Fixes #1196

## Checklist

## Screenshots

Image 2020-04-27 at 12 30 58 PM


app/views/demos/show.html.erb
```diff
@@ -16,12 +16,20 @@
...
             data: {action: "change->ad-demo#update", controller: "select", target: "ad-demo.themeSelect"}}) %>
     </div>
+    <% if current_user && current_user.campaigns.present? %>
+      <div class="col-md-6 offset-md-3">
+        <label class="form-label select" for="ad_demo_theme">Campaign:</label>
+        <%= form.collection_select(:campaign_id, current_user.campaigns, :id, :to_s, {selected: @campaign_id}, {
+              class: "form-control custom-select", id: "ad_demo_campaign_id",
+              data: {action: "change->ad-demo#update", controller: "select", target: "ad-demo.campaignIdSelect"}}) %>
+      </div>
+    <% end %>
```
hopsoft, line 32: Should we omit `archived` campaigns?

coderberry: I think we should leave it. It would be good for advertisers to be able to review them.

PR #1240, opened 2020-04-30
cards, list groups, and tabs

andrewmcodes:
## Type of PR

- [x] feature
- [ ] enhancement
- [ ] bug fix
- [ ] other

## Description

- Creates cards, list group, and tabs components.
- Removes websites controller and resources since it's unused.

## Checklist


andrewmcodes:
Not sure why system tests are failing in CI - they pass locally. This happens occasionally for some reason.

app/components/avatar_component.rb
```diff
@@ -0,0 +1,25 @@
+class AvatarComponent < ViewComponent::Base
```
coderberry, line 1: This prevents me from setting a specific size for the gravatar. Also, if we are using the gravatar, do we want to pass in the size (converted to px) in the `user_gravatar_url` call?

app/components/card_component.rb
```diff
@@ -0,0 +1,19 @@
...
+    @class_names = classes
+  end
+
+  private
+
+  attr_reader :class_names
+
+  def classes
+    classes = ["card"]
+    classes.push(class_names) unless class_names.try(:empty?)
```
coderberry, line 12:
```suggestion
    classes.push(class_names) unless class_names.empty?
```

coderberry: I think you can call .empty? on nil right?

andrewmcodes: You cannot :)

app/components/avatar_component.rb
andrewmcodes: We are already using this today, I just took it out of the module it was in. If you’d like to look into that implementation, we should do it in another PR

app/components/card_component.rb
andrewmcodes: But I can probably refactor this now that I look at it

PR #1251, opened 2020-05-05
Inbound Email parsing

coderberry:
## Type of PR

- [x] feature
- [ ] enhancement
- [ ] bug fix
- [ ] other

## Description

Allow capture of inbound emails to CodeFund team members from either advertisers or publishers and stores emails.

This is phase 1 of migrating out of Pipedrive into CF for account management.

Note: For this to work on stage/prod, we must set the ENV var `RAILS_INBOUND_EMAIL_PASSWORD` and then run `rails misc:generate_inbound_destination_url`. This URL is then used by SendGrid for the ingress email settings.

## Checklist


app/views/user_emails/_email_preview.erb
```diff
@@ -0,0 +1,20 @@
+<div class="list-group list-group-messages list-group-flush list-group-bordered">
```
andrewmcodes, line 1: Lets use `ListGroupComponent` here. You can just add another arg called `messages` and follow the same pattern as the others. Lmk if you want to pair on this and knock it out.

app/views/user_emails/_email_preview.erb
```diff
@@ -0,0 +1,20 @@
...
+        <div class="col-12 col-lg-9">
+          <h4 class="list-group-item-title text-truncate">
+            <%= link_to email.subject, user_email_path(current_user, email) %>
+          </h4>
+          <p class="list-group-item-text text-truncate"><%= email.snippet %></p>
+        </div>
+      </div>
+    </div>
+  </div>
+</div>
```
andrewmcodes, line 20: Add newline

app/views/user_emails/show.html.erb
```diff
@@ -0,0 +1,31 @@
...
+          subtitle: @user.email,
+          back_link: BackLinkComponent.new(title: "Emails", link: user_emails_path(current_user)),
+          sidebar: true, buttons: [
+            layout_button(link: edit_user_path(@user), icon: "fad fa-pencil", title: "Edit user"),
+            layout_button(link: user_impersonation_path(@user), method: :put, icon: "fad fa-mask", title: "Masquerade as user", admin: true)
+          ])) %>
+  <% end %>
+  <% component.with(:body) do %>
+    <%= render CardComponent.new(classes: "trix-email-preview") do %>
+      <div class="list-group list-group-messages list-group-flush list-group-bordered">
```
andrewmcodes, line 14: Same comment as above for using `ListGroupComponent`

bin/copy_production_db_to_local
```diff
@@ -2,7 +2,7 @@
 source .env
 export RAILS_ENV=development
 bundle exec rails db:migrate db:drop db:create
-./bin/heroku_pg_dump_production_replica
+bash ./bin/heroku_pg_dump_production_replica
 pg_restore -U $PG_USERNAME --verbose --clean --no-acl --no-owner -h localhost -d code_fund_ads_development tmp/production-shallow.dump
 bundle exec rails db:migrate
-./bin/generate_db_artifacts
+bash ./bin/generate_db_artifacts
```
andrewmcodes, line 8: Do you need to add bash here since both of these binstubs have `#!/usr/bin/env bash` at the top?

config/environments/development.rb
```diff
@@ -1,6 +1,9 @@
 require "dotenv/load"
 
 Rails.application.configure do
+  # Prepare the ingress controller used to receive mail
+  config.action_mailbox.ingress = :sendgrid
```
andrewmcodes, line 5: Should we be using SendGrid in development? Answer may be yes - just wanted to make sure.

lib/tasks/misc.rake
```diff
@@ -0,0 +1,6 @@
+namespace :misc do
```
andrewmcodes, line 1: maybe `mailer` is a better name? Just a nitpick - feel free to ignore

test/components/back_link_component_test.rb
```diff
@@ -0,0 +1,10 @@
+require "test_helper"
+
+class BackLinkComponentTest < ViewComponent::TestCase
+  test "component renders" do
+    result = render_inline(BackLinkComponent.new(title: "TITLE", link: "LINK"))
+    assert_selector("ol.breadcrumb")
+    assert_selector("a[href=LINK]")
+    assert_includes result.css("a[href=LINK]").to_html, "TITLE"
+  end
```
andrewmcodes, line 9: 🎉

app/components/back_link_component.rb
```diff
@@ -0,0 +1,6 @@
+class BackLinkComponent < ApplicationComponent
+  def initialize(title:, link:)
```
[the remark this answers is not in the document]
coderberry: By having this, you are allowing nil to be a valid value. We want it to not work unless there's values passed. Are there reasons to keep it nil?

app/components/back_link_component.html.erb
```diff
@@ -0,0 +1,9 @@
+<nav aria-label="breadcrumb">
+  <ol class="breadcrumb">
+    <li class="breadcrumb-item active">
+      <%= link_to @link do %>
+        <i class="breadcrumb-icon fa fa-angle-left mr-2"></i><%= @title %>
```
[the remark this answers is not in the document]
coderberry: I'm curious to the benefits of doing it this way. Is it more a style preference?

app/models/inbound_email.rb
```diff
@@ -0,0 +1,48 @@
...
+  # callbacks .................................................................
+  # scopes ....................................................................
+
+  # additional config (i.e. accepts_nested_attribute_for etc...) ..............
+  has_rich_text :body
+  has_many_attached :attachments
+
+  # class methods .............................................................
+  class << self
+  end
```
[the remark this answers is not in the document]
coderberry: Will do. However, I think this is how most of our models are configured.

bin/copy_production_db_to_local
coderberry: probably not. I'll remove them

config/environments/development.rb
coderberry: the ingress can be configured without it actually bringing in emails. I think this is optional, but doesn't hurt and allows for easier testing locally.

lib/tasks/misc.rake
coderberry: I'd rather leave it as misc because this command will be rarely run. I like to keep my one-time runs in a misc namespace to denote the lack of importance.

app/components/back_link_component.rb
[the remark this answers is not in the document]
andrewmcodes: If you default to nil, and use the `render?` method, you let the component be in charge of choosing to render or not. This way if someone screws up and it slips to production, it will simply just not render rather than causing an error that prevents the page from loading. Something I’ve also been considering is raising in render only in development if they aren’t passed in. 

```ruby
def render?
  raise ArgumentError.new("Title and link cannot be nil") if Rails.env.development && (title.nil? || link.nil?)
  title && link
end 
```

app/models/inbound_email.rb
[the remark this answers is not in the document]
andrewmcodes: Your choice!

app/components/back_link_component.html.erb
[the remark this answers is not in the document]
andrewmcodes: Yes and also just being very specific about what’s available to avoid any unexpected visitors. It’s the same idea as passing variables down to your partials.

app/views/user_emails/show.html.erb
```diff
@@ -0,0 +1,31 @@
...
+          subtitle: @user.email,
+          back_link: BackLinkComponent.new(title: "Emails", link: user_emails_path(current_user)),
+          sidebar: true, buttons: [
+            layout_button(link: edit_user_path(@user), icon: "fad fa-pencil", title: "Edit user"),
+            layout_button(link: user_impersonation_path(@user), method: :put, icon: "fad fa-mask", title: "Masquerade as user", admin: true)
+          ])) %>
+  <% end %>
+  <% component.with(:body) do %>
+    <%= render CardComponent.new(classes: "trix-email-preview") do %>
+      <%= render ListGroupComponent.new(classes: "list-group-messages list-group-flush list-group-bordered") do %>
```
andrewmcodes, line 14: This is one way of doing it - though `ListGroupComponent` has some args to make this easier. See [the class initializer](https://github.com/gitcoinco/code_fund_ads/blob/master/app/components/list_group_component.rb#L2)

`<%= render ListGroupComponent.new(bordered: true, flush: true) do %>`

I would suggest just adding messages as another named arg to the component so the component would then look like:

```ruby
class ListGroupComponent < ApplicationComponent
  def initialize(bordered: false, flush: false, reflow: false, messages: false, classes: nil)
    @bordered = bordered
    @flush = flush
    @reflow = reflow
    @messages = messages
    @class_names = classes
  end

  private

  attr_reader :class_names, :bordered, :flush, :reflow, :messages

  def classes
    classes = ["list-group"]
    classes << "list-group-bordered" if bordered
    classes << "list-group-flush" if flush
    classes << "list-group-reflow" if reflow
    classes << "list-group-messages" if messages
    classes << class_names if class_names
    classes.compact
  end

  def render?
    content.present?
  end
end
```

Since we are using the messages class in more than one place I would recommend doing it this way, which would make your render `<%= render ListGroupComponent.new(bordered: true, flush: true, messages: true) do %>` but you could just not add messages and do `<%= render ListGroupComponent.new(bordered: true, flush: true, classes: "list-group-messages") do %>`. Reducing the use of the `classes` escape hatch will make updating styles easier though since there is a single source of truth.

app/views/user_emails/show.html.erb
```diff
@@ -0,0 +1,31 @@
...
+              <%= @email.sender %>
+            </h4>
+            <p class="list-group-item-text"><%= @email.delivered_at %></p>
+          </div>
+        </div>
+        <div class="list-group-item unread">
+          <div class="list-group-item-body pl-md-2">
+            <%= @email.body %>
+          </div>
+        </div>
```
andrewmcodes, line 27:
```suggestion
        <%= render ListGroup::ItemComponent.new(classes: "unread") do %>
          <div class="list-group-item-body pl-md-2">
            <h4 class="list-group-item-title text-truncate">
              <%= @email.sender %>
            </h4>
            <p class="list-group-item-text"><%= @email.delivered_at %></p>
          </div>
        </div>
        <div class="list-group-item unread">
          <div class="list-group-item-body pl-md-2">
            <%= @email.body %>
          </div>
        <% end %>
```

app/views/user_emails/_email_preview.erb
andrewmcodes, line 1: If you apply my suggestion below this would become 

```ruby
<%= render ListGroupComponent.new(bordered: true, flush: true, messages: true) do %>
  <%= render ListGroup::ItemComponent.new(classes: "unread") do %>
    <div class="list-group-item-body pl-md-2">
      <div class="row">
        <div class="col-12 col-lg-3 d-none d-lg-block">
          <h4 class="list-group-item-title text-truncate">
            <%= link_to email.sender, user_email_path(current_user, email) %>
          </h4>
          <p class="list-group-item-text"><%= time_ago_in_words email.delivered_at %> ago</p>
        </div>
        <div class="col-12 col-lg-9">
          <h4 class="list-group-item-title text-truncate">
            <%= link_to email.subject, user_email_path(current_user, email) %>
          </h4>
          <p class="list-group-item-text text-truncate"><%= email.snippet %></p>
        </div>
      </div>
    </div>
  <% end %>
<% end %>
```

app/views/user_emails/show.html.erb
coderberry: You realize how much complexity is being added to a simple div right?

andrewmcodes: Also if you add `messages`, this would be a great test to add to `test/components/list_group_component_test.rb`

```ruby
  test "messages list group" do
    assert_equal(
      %(<ul class="list-group list-group-messages"> List Group </ul>),
      render_inline(ListGroupComponent.new(messages: true)) { "List Group" }.to_html.squish
    )
  end
```

PR #1295, opened 2020-05-20
Bug fix - User limit to their own campaigns

### LINKED ISSUE #1284: Preview Ads does not work after switching to another template/theme
# Bug Report

## Describe the bug

See Steps to Reproduce.

## Expected behavior

We should only allow the user to view *their* ads even when switching template/themes.

## Steps To Reproduce

1. Go to any campaign
2. Click 'Preview Ads'
3. Note that the ad being previewed is from the campaign I was on
4. Select another ad template
5. Note that the ad displayed is no longer the correct one.

coderberry:
## Type of PR

- [ ] feature
- [ ] enhancement
- [x] bug fix
- [ ] other

## Description

Added campaigns to be scoped to the current user's viewable campaigns. Also updated the style a bit.

Fixes #1284 

## Checklist

## Screenshots

Image 2020-05-20 at 4 11 49 PM


bin/copy_production_db_to_local
```diff
@@ -3,7 +3,7 @@
 source .env
 export RAILS_ENV=development
 
-bundle exec rails db:migrate db:drop db:create
+bundle exec rails db:drop db:create
```
andrewmcodes, line 6: I’m not sure this change is desired - what prompted you to change it?

app/controllers/demos_controller.rb
```diff
@@ -29,4 +31,14 @@ def campaign
...
     params.permit(:template, :theme, :campaign_id)
   end
+
+  def load_campaigns
+    @campaigns = if authorized_user&.can_admin_system?
+      Campaign.all.order(name: :asc)
+    elsif current_user&.campaigns
+      current_user.campaigns.order(name: :asc)
+    else
+      []
```
andrewmcodes, line 41: One of the original goals of this demo page was so that everyone - publishers, potential advertisers, etc. - would be able to see our ads in action with all of the available templates. That’s why it was scoped to only be querying a specific fallback campaign we setup specific for this. 

I just want to make sure that you’re intending to remove this feature for those users. If that’s the case, we should put some authorizations in place so only current advertisers can see this page.

bin/copy_production_db_to_local
coderberry: By running migrate right before drop, the drop can hang b/c the database is in use. There's no reason to migrate right before dropping a database.

app/controllers/demos_controller.rb
coderberry: I agree. This still works great if you are not authenticated. If you *are* authenticated, it will allow you to select the campaigns from your list of campaigns (if available) and if not, it will fall back to the default.

andrewmcodes: Ahhh ok I read this wrong at first bc I’m on mobile and GitHub folded part of the code which made arrive at that incorrect understanding of this change. Disregard :) 

However do you need a break on 37? It’s possible for an admin to meet the if and elsif condition (admin who has also created fallback campaigns). Could be wrong but something about this is definitely making me suspicious

PR #1304, opened 2020-05-21
Email threading, cleanup and inbox

coderberry:
## Type of PR

- [x] feature
- [ ] enhancement
- [ ] bug fix
- [ ] other

## Checklist


app/javascript/themes/current/stylesheets/components/page.scss
```diff
@@ -0,0 +1,29 @@
...
+    overflow-y: auto;
+  }
+  .page-fixed-footer {
+    position: relative;
+    padding: 0.5rem;
+    background-color: #fff;
+    box-shadow: 0 -1px 0 0 rgba(20, 20, 31, 0.075);
+    z-index: 1;
+  }
+}
```
andrewmcodes, line 29: Curious why you had to add this. Did looper not have what you needed?

app/javascript/themes/current/stylesheets/components.scss
```diff
@@ -2,9 +2,10 @@
 @import './components/checkbox_tree';
 @import './components/creative_preview';
 @import './components/dropdown';
+@import './components/emails';
```
andrewmcodes, line 5: I don't see this file.

app/models/concerns/users/presentable.rb
```diff
@@ -20,8 +20,8 @@ def hashed_email
       Digest::MD5.hexdigest(email.downcase)
     end
 
-    def gravatar_url(d = "identicon")
-      "https://www.gravatar.com/avatar/#{hashed_email}?s=300&d=#{d}"
+    def gravatar_url(d = "identicon", s = "300")
```
andrewmcodes, line 23: Did you use this change anywhere? If not I'd argue we revert it but up to you.

app/views/shared/_sidebar.html.erb
```diff
@@ -34,7 +34,7 @@
         <ul class="menu">
           <%= render "/shared/sidebar_content/dashboard" unless authorized_user.can_advertise_and_publish? %>
           <%= render "/shared/sidebar_content/dashboards" if authorized_user.can_advertise_and_publish? %>
-          <%#= render "/shared/sidebar_content/inbox" if authorized_user.can_view_emails? %>
+          <%= render "/shared/sidebar_content/inbox" if authorized_user.can_view_emails? %>
```
andrewmcodes, line 37: should this only be for admins?

config/initializers/closure_tree_config.rb
```diff
@@ -0,0 +1,5 @@
+ClosureTree.configure do |config|
+  # Some PaaS like Heroku don't have available the db in some build steps like
+  # assets:precompile, this is skipped when this value is true, default = false
+  # config.database_less = true
```
andrewmcodes, line 4: Do we need to uncomment this based on the comment above?

Gemfile
```diff
@@ -72,6 +73,7 @@ gem "simple_form", "~> 5.0"
 gem "slack-notifier", "~> 2.3.2"
 gem "split", "~> 3.4.1", require: "split/dashboard"
 gem "spreadsheet", "~> 1.2.4"
+gem "unread", "~> 0.11.0"
```
andrewmcodes, line 76: **TL;DR I know we need to get this out the door so I won't hold this PR up bc of this.**

But..

Im not really sure we get a lot of benefit out of this gem vs doing this ourselves and I am worried locking us in to it will only create problems down the road. The fact we are only _really_ using it to mark emails as read and that there is no builtin way to mark emails as unread makes me even more convinced we shouldn't add this. I don't think it would take more than an hour or two to replace it if that's the concern. 

Happy to chat more about this.

coderberry: I like your feedback and if it you want we can replace it. The criteria we will need though really fits the gem. I'll need to know who read the messages and search by that stat. To me it looked like a well maintained gem that solved a need quickly. Using gems is why devving in Ruby on Rails is 10x right?

app/javascript/themes/current/stylesheets/components.scss
coderberry: I see it

app/javascript/themes/current/stylesheets/components/page.scss
coderberry: That is correct. Looper didn't have what I needed there (scrollable middle with fixed top and bottom)

app/models/concerns/users/presentable.rb
coderberry: I was going to use it but I didn't. This change does allow for better sizing though so there's no reason not to keep it imho.

app/views/shared/_sidebar.html.erb
coderberry: It's only for authorized users that can view emails. I did it this way so we can add more rules in the future.

config/initializers/closure_tree_config.rb
coderberry: Possibly

app/views/shared/_sidebar.html.erb
andrewmcodes: Ah didn't see that we already had an authorizer.

app/javascript/themes/current/stylesheets/components.scss
andrewmcodes: Oh it's already in the app - my b.

db/migrate/20200521213149_add_email_threading_to_emails.rb
```diff
@@ -0,0 +1,8 @@
+class AddEmailThreadingToEmails < ActiveRecord::Migration[6.0]
+  def change
+    add_column :emails, :in_reply_to, :string
+    add_column :emails, :message_id, :string
+    add_column :emails, :parent_id, :bigint
+    add_index :emails, :parent_id
```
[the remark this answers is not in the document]
andrewmcodes: Just add lol_dba and see what it says

PR #1310, opened 2020-05-27
Conversion Pixel

coderberry:
## Type of PR

- [x] feature
- [ ] enhancement
- [ ] bug fix
- [ ] other

## Description

https://jsbin.com/dawagiyeda/edit?html,console,output

## Checklist

- [ ] My code follows the style guidelines of this project & the tests are passing
- [ ] I have performed a self-review of my own code
- [ ] I have requested a review from another developer(s)


app/javascript/src/conversion-tracker.js
```diff
@@ -28,9 +28,10 @@ export default class {
 
   // Indicates if the passed date (represented as a string from localStorage) has expired based on daysToLive
   expired (dateString) {
-    const date = Date.parse(dateString)
-    const expiresAt = new Date(date.getTime())
-    expiresAt.setDate(expiresAt.getDate() + this.daysToLive)
+    const date = new Date(Date.parse(dateString)) // => Tue Jun 09 2020 14:33:59 GMT-0400 (EDT)
+    let expiresAt = new Date()
+    expiresAt.setTime(date.getTime())
+    const setAt = expiresAt.setDate(expiresAt.getDate() + this.daysToLive)
```
andrewmcodes, line 34: This needs a look (and the comment can be removed) but I am pretty sure this is what we want. We may be able to make it a bit cleaner but this works.

app/models/pixel.rb
```diff
@@ -49,7 +49,7 @@ class << self
   # Records an conversion for the passed impression_id
   # This operation is a noop if no impression is found
   def record_conversion(impression_id, test: false, metadata: {})
-    impression = Impression.find_by(id: impression_id, campaign_id: Campaign.where(organization_id: organization_id).select(:id))
```
andrewmcodes: I really wasn't sure the decision behind finding the impression this way so there is a complete possibility that my change is not correct. However, this does work to correctly pull the impression id and org id and find the impression.

config/routes.rb
```diff
@@ -23,7 +23,7 @@
 
   # resource :global_stats, only: [:show], defaults: {format: :json}
 
-  match "/pixels/:pixel_id/convert/:impression_id", to: "pixel_conversions#create", via: [:get, :post], as: "pixel_conversions"
+  match "/pixels/:pixel_id/impressions/:impression_id", to: "pixel_conversions#create", via: [:get, :post], as: "pixel_conversions"
```
andrewmcodes, line 26: Nate used impressions in the XHR request which was different from the route specified but it made more sense to me to have the route structured this way so I modified the route instead of the fetch.

PR #1313, opened 2020-05-31
Makes session works without caching-dev.txt

woto:
## Type of PR

- [ ] feature
- [ ] enhancement
- [x] bug fix
- [ ] other

## Description

[In case of absence of `caching-dev.txt`](https://github.com/gitcoinco/code_fund_ads/blob/master/config/environments/development.rb#L24)

[cache_store initializes as :null_store](https://github.com/gitcoinco/code_fund_ads/blob/master/config/environments/development.rb#L39)

[and later when it uses as session store](https://github.com/gitcoinco/code_fund_ads/blob/master/config/initializers/sessions.rb#L3)

it simply disables Rails `session` functionality. As a result there is no way to do almost all actions which relies on Rails security csrf mechanism.

## Checklist

### DIFF
--- a/config/initializers/sessions.rb
+++ b/config/initializers/sessions.rb
@@ -1,5 +1,7 @@
 # Using redis to manage sessions (same cache appliance as standard rails caching)
 # IMPORTANT: "expire_after" should match "timeout_in" config in: config/initializers/devise.rb
-Rails.application.config.session_store ActionDispatch::Session::CacheStore,
-  key: "_code_fund_ads_session",
-  expire_after: 2.weeks
+unless Rails.application.config.cache_store == :null_store
+  Rails.application.config.session_store ActionDispatch::Session::CacheStore,
+    key: "_code_fund_ads_session",
+    expire_after: 2.weeks
+end


coderberry:
@hopsoft can you take a look at this when you get a chance?

andrewmcodes:
@coderberry I'll take a look - need to test a few things

woto:
@andrewmcodes No, all is correct. But as a usual developer, I didn't read instructions (joking). I read it, but successfully ignored step:

> bundle exec rails dev:cache

as I perceived this step as optional.

> It is recommended to develop with Rails cache enabled.

> This application relies heavily on caching and may not work properly without the cache enabled.

And then I delved into the guts of the ActionController::RequestForgeryProtection to realize why I'm unable to login. At least with this change, we can log in to backend in the opposite of nothing. I think if you are against this change then this installation step in documentation should be changed to obligatory.  Agree with me?

andrewmcodes:
@all-contributors please add @woto for code

woto:
@andrewmcodes Cool, thanks. I'm Looking forward weekends to discovering about this project more, because of interests in this area

